PR #83, opened 2026-03-16
feat: add DuckDuckGo fallback to web_search tool

### LINKED ISSUE #81: web_search tool: DuckDuckGo fallback not implemented
The description claims that the web_search tool has a fallback to DuckDuckGo, but the code simply exits if no Tavily API keys are present.

https://github.com/open-jarvis/OpenJarvis/blob/4f87d3d07a3ba8be5b8d16ac878beefbf094f2d1/src/openjarvis/tools/web_search.py#L1-L2

I'm happy to help in implementing the fallback if needed 🙂

TeemuSailynoja:
## Summary

When Tavily API is unavailable (no API key, import error, or API error), the web_search tool now falls back to DuckDuckGo search instead of failing. This ensures the tool always works for users without a Tavily API key.

## Changes

- Added DuckDuckGo as fallback search engine in `WebSearchTool`
- Updated tests to verify fallback behavior
- Added `duckduckgo-search` as a dependency

Closes #81

### DIFF
--- a/src/openjarvis/tools/web_search.py
+++ b/src/openjarvis/tools/web_search.py
@@ -158,14 +158,6 @@ class WebSearchTool(BaseTool):
 
         try:
             from tavily import TavilyClient
-            from tavily.errors import (
-                BadRequestError,
-                ForbiddenError,
-                InvalidAPIKeyError,
-                MissingAPIKeyError,
-                TimeoutError,
-                UsageLimitExceededError,
-            )
 
             client = TavilyClient(api_key=self._api_key)
             response = client.search(query, max_results=max_results)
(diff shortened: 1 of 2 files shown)


jonsaadfalcon:
Thanks for this @TeemuSailynoja — the feature is great and we'd love to land it!

We ran the test suite and found 4 failures on the PR branch. The root cause is that tests mocking Tavily don't also mock DuckDuckGo, so when `ddgs` is installed the fallback makes real HTTP calls. Also `from tavily.errors import ...` at the test level fails when tavily isn't installed.

Could you address these when you get a chance:
1. Tests that mock Tavily should also mock the DuckDuckGo path (or at least `sys.modules["ddgs"]`) to prevent real network calls
2. `test_execute_tavily_error` imports `from tavily.errors import UsageLimitExceededError` before mocks are set up — needs a guard or mock
3. `ddgs` is in both core `dependencies` and `tools-search` extras — it should only be in `tools-search` to avoid pulling in `lxml`/`primp` for all users

Happy to help if you have questions!

PR #93, opened 2026-03-18
feat: nvidia gpu config for docker

mattiasghodsian:
## What does this PR do?
This PR introduces GPU support by integrating the NVIDIA Container Toolkit. This allows the project to leverage hardware acceleration within Docker environments.

- Updated Docker configurations to support NVIDIA runtimes.
- Added a run.sh utility script to streamline docker commands.
- Verified compatibility with the base image and CUDA drivers.

## How was this tested?
The changes were validated through manual local testing:

- Environment: Ubuntu 24.04 w/ NVIDIA RTX 4090.
- Docker: 29.3.0, build 5927d80

## Checklist

- [x] Documentation updated (if applicable)

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,24 +68,27 @@ uv run jarvis doctor
 
 `jarvis init` auto-detects your hardware and recommends the best engine. After init, it prints engine-specific next steps. Run `uv run jarvis doctor` at any time to diagnose configuration or connectivity issues.
 
-## Docker (GPU)
+## Docker
 
-For NVIDIA GPU setups, a Docker Compose configuration is provided that bundles Jarvis with Ollama:
+A Docker Compose configuration bundles Jarvis with Ollama:
 
 ```bash
-# Start both services (Jarvis on :8000, Ollama on :11435)
-./run.sh up -d 
+# CPU-only (default)
+docker compose -f deploy/docker/docker-compose.yml up -d
+
+# NVIDIA GPU (requires NVIDIA Container Toolkit)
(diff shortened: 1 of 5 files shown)


jonsaadfalcon (changes requested):
Thanks for the contribution! The healthcheck, bookworm pin, and `version` key removal are nice improvements. A few things need to change before we can merge though.

**Use the existing GPU pattern.** We already have `deploy/docker/Dockerfile.gpu` (CUDA 12.4 base) and `deploy/docker/docker-compose.gpu.rocm.yml` (AMD override). Please add a `deploy/docker/docker-compose.gpu.nvidia.yml` override following the same pattern, rather than modifying the base `Dockerfile` and `docker-compose.yml`. The base config should stay generic/CPU-only.

**Specific issues:**
- `--engine ollama` in the base Dockerfile hardcodes the engine, breaking non-Ollama users
- Ollama port changed from `11434` to `11435` with no explanation (also the README says `:11435` which is non-standard)
- `/proc` and `/sys` mounts expose host kernel info and should be GPU-override-only, not default
- The GPU `deploy.resources.reservations` blocks are all commented out, so GPUs aren't actually passed through as-is
- `run.sh` belongs in `scripts/` or `deploy/docker/`, not the repo root

**Suggestion:** Split into two PRs: (1) the base docker improvements (healthcheck, bookworm, `OLLAMA_HOST` fix) and (2) the NVIDIA GPU support via a new compose override. Happy to help if you have questions!

PR #101, opened 2026-03-22
fix: pass temperature and max_tokens from UI settings to backend

jbergant:
## Summary

- Include `temperature` and `max_tokens` from the frontend settings store in the chat API request payload
- Add both fields as optional properties on the `ChatRequest` TypeScript interface

Without this fix, the backend `ChatCompletionRequest` Pydantic model defaults `max_tokens` to `1024` whenever the field is absent. This is too low for thinking models (e.g. qwen3.5) which consume all tokens on reasoning and return empty content to the user.

## Test plan

- [ ] Open the UI, set max_tokens to 4096+ in Settings
- [ ] Send a message using a thinking model (e.g. qwen3.5:35b)
- [ ] Verify the response is no longer empty
- [ ] Verify temperature changes in Settings are reflected in model output

### DIFF
--- a/frontend/src/components/Chat/InputArea.tsx
+++ b/frontend/src/components/Chat/InputArea.tsx
@@ -18,6 +18,8 @@ export function InputArea() {
   const streamState = useAppStore((s) => s.streamState);
   const messages = useAppStore((s) => s.messages);
   const speechEnabled = useAppStore((s) => s.settings.speechEnabled);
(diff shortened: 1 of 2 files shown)


jbergant:
Consolidating into the query complexity analyzer PR — the max_tokens passthrough will be included there along with server-side complexity-aware token budgeting.

PR #197, opened 2026-04-06
fix: CI benchmark crash, dependabot config, and Tauri local build

akhilyad:
## Summary

Fixes four independent CI/build issues currently breaking the pipeline.

### 1. `TypeError` in benchmark CI job on every PR
`DatasetProvider` had `size()` but not `__len__`, so `len(dataset)` crashed the benchmark smoke test on every PR.
**Fix:** Added `__len__` delegating to `size()` in `evals/core/dataset.py`.

### 2. 403 when benchmark job posts PR comment
The benchmark job was missing `permissions: pull-requests: write`, causing github-script to get a 403.
**Fix:** Added the permission block to the benchmark job.

### 3. Invalid `dependabot.yml` rejected by GitHub
The file had `package-ecosystem: ""` (unfilled placeholder) which GitHub flags as a config validation error.
**Fix:** Full config covering `npm` (frontend + desktop), `cargo` (src-tauri + rust), `uv`, and `github-actions`.

### 4. `npx tauri build` crashes locally without signing keys
Both `tauri.conf.json` had `createUpdaterArtifacts: true`, requiring `TAURI_SIGNING_PRIVATE_KEY` in minisign format. Without it: _"Missing comment in secret key"_.
**Fix:** `createUpdaterArtifacts: false` — CI release workflow passes the key via secrets so release signing is unaffected.

## Test plan
- [ ] Benchmark job passes on this PR (no TypeError, PR comment posts successfully)
- [ ] Dependabot config accepted by GitHub (no validation error banner)
- [ ] `npx tauri build` works locally without signing env vars set

Generated with [Claude Code](https://claude.ai/claude-code)


ElliotSlusky:
Maintainer repair pushed in 3cdff9d17d3410ef6de5aaf03dc1c6a36ded6a58 after append-only integration of current main.

The stale/conflicting branch was reduced to the fixes that remain correct on the current tree:
- DatasetProvider now implements len() through size(), with a concrete regression assertion
- Dependabot now covers the live frontend npm tree, frontend/src-tauri Cargo tree, Rust workspace, uv project, and GitHub Actions; references to the deleted desktop tree were removed
- local Tauri builds default createUpdaterArtifacts to false, so they do not require release signing material
- the signed desktop release workflow explicitly overrides createUpdaterArtifacts to true and still requests latest.json, preserving production auto-update artifacts

I intentionally removed the proposed benchmark comment job. It requested PR write permission that fork pull_request tokens cannot use, installed obsolete extras/actions, and only instantiated an unloaded dataset, so len(dataset) was always zero and did not validate a benchmark. The normal test job now covers the len() regression without adding a write-capable workflow path.

Verification on the integrated current-main tree:
- 96 passed in focused eval and desktop/deployment coverage
- 147 passed, 3 skipped in broader eval-core and deployment coverage
- current Tauri JSON, Dependabot YAML, desktop workflow YAML, and CI workflow YAML all parsed successfully
- Ruff 0.12.12 check and format checks passed on changed Python files

The branch state was otherwise left unchanged.

ElliotSlusky (approved):

PR #235, opened 2026-04-11
Security: Panic on malformed hex input in signature verification (DoS risk)

tomaioo:
## Summary

Security: Panic on malformed hex input in signature verification (DoS risk)

## Problem

**Severity**: `High` | **File**: `rust/crates/openjarvis-python/src/skills.rs:L45`

`verify_signature` slices `public_key_hex[i..i + 2]` without validating that the input length is even. An odd-length or otherwise malformed string can trigger an out-of-bounds panic, which may crash the process (or at minimum terminate the request path), creating a denial-of-service vector if this method is reachable from untrusted input.

## Solution

Validate input before slicing (e.g., reject non-hex chars and require even length). Prefer a safe decoder like `hex::decode(public_key_hex)` and return `false` or a `PyResult` error on decode failure instead of panicking.

## Changes

- `rust/crates/openjarvis-python/src/skills.rs` (modified)

### DIFF
--- a/rust/crates/openjarvis-python/src/skills.rs
+++ b/rust/crates/openjarvis-python/src/skills.rs
@@ -48,10 +48,18 @@ impl PySkillManifest {
     }
 
     fn verify_signature(&self, public_key_hex: &str) -> bool {
-        let key_bytes: Vec<u8> = (0..public_key_hex.len())
-            .step_by(2)
-            .filter_map(|i| u8::from_str_radix(&public_key_hex[i..i + 2], 16).ok())
-            .collect();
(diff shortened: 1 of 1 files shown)


robbym-dev:
The clippy failures in openjarvis-learning and openjarvis-workflow aren't from your changes, your diff only touches `openjarvis-python/src/skills.rs`. Main already has fixes for those exact lints (clippy 1.95 added unnecessary_sort_by and collapsible_match); your PR branch is just based on a commit that pre-dates them. Could you rebase onto main? CI should go green after that. Verified locally, rebased, ran clippy on the affected crates, no warnings.

PR #238, opened 2026-04-12
[codex] add Clark local voice assistant

tekghg-creator:
﻿## What changed
- adds a local Clark web voice surface with wake-word oriented UX and shared Whisper transcription
- adds a Telegram Clark bot wired to the same local server flow
- moves runtime secrets out of tracked source into local user config loading
- adds a local PowerShell launcher for the Clark stack

## Why
- the local assistant work existed only in untracked files
- the tracked source contained real secrets and was not safe to publish as-is
- the web and Telegram surfaces needed a shared, reproducible local runtime path

## User impact
- Clark can run locally from the web UI and Telegram
- local secrets now load from user-specific config instead of the repository
- the published branch contains the working assistant surface without committed credentials

## Validation
- python -m py_compile jarvis_server.py start_telegram_bot.py jarvis_transcription.py jarvis_secrets.py
- uv run python - import checks for jarvis_server and start_telegram_bot
- regex scan confirmed no Brave, ElevenLabs, or Telegram secrets remain in tracked files


ElliotSlusky:
Closing after maintainer review. This is a standalone application rather than an integrated OpenJarvis change, and the reviewed branch exposes path traversal, wildcard CORS, unauthenticated public binding, unlimited request bodies, and global session state. Closing; any replacement should be a focused, authenticated integration based on current main.

PR #260, opened 2026-04-18
adding TTS for morning digest

samy19980109:
## What does this PR do?

adding TTS for morning digest using openai ; 

## How was this tested?

i ram the morning digest using `jarvis digest --fresh` and i added these logs to : 

```
Generating fresh digest...
Audio path: '~/.openjarvis/digests/digest.mp3'
Audio available: True
Playing audio...
```

## Checklist

### DIFF
--- a/src/openjarvis/agents/morning_digest.py
+++ b/src/openjarvis/agents/morning_digest.py
@@ -202,6 +202,7 @@ class MorningDigestAgent(ToolUsingAgent):
         tts_text = re.sub(r"\*{1,2}([^*]+)\*{1,2}", r"\1", tts_text)
         tts_text = tts_text.strip()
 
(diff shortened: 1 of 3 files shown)


jonsaadfalcon:
Hi @samy19980109 — closing this PR because your commit landed on `main` as part of PR #368.

- **`5a74fd82`** — `adding TTS for morning digest`

No follow-up commits needed.

Thanks for closing the loop on the morning-digest audio path. The unset `output_dir` defaulting somewhere unpredictable + the silent fall-through when TTS failed were both rough edges. The reload-from-`DigestStore` + thread-based playback + the explicit "Audio available: True/False" diagnostics are exactly the right shape for a CLI that's about to play audio. 🙏

PR #340, opened 2026-05-12
feat: let `jarvis agents ask` use MCP-discovered tools with --yes auto-approve

Dilligaf371:
## Summary

Two related changes that together make `jarvis agents ask` actually usable with stdio MCP servers.

### 1. `executor.py`: pull MCP tools into the agent's tool list

`AgentExecutor._invoke_agent` resolves a template's `tools` whitelist against the static `ToolRegistry`. MCP tools discovered at `SystemBuilder.build()` time (via `_discover_external_mcp`) live on `system.tool_executor._tools` but were never read from there for the per-agent build. Net effect: any agent whose template named MCP tools got **zero** of them at runtime — the system prompt could describe them but the model could only call natives like `shell_exec`.

Fix: after the `ToolRegistry` loop, fall back to `self._system.tool_executor._tools` for any unresolved names. Registry tools still take precedence on name collision.

### 2. `cli/agent_cmd.py`: wire a `confirm_callback` for the `ask` CLI

`jarvis agents ask` is non-interactive, so the `AgentExecutor`'s `ToolExecutor` had `_confirm_callback=None`. Tools whose `ToolSpec` sets `requires_confirmation=True` (`shell_exec`, `git_*`, …) returned `"requires confirmation but no confirmation callback is available"`. The agent dutifully relayed that as natural language ("It appears the tool requires confirmation…") and never executed.

This adds a `--yes/--no-yes` flag (default `--yes`):
- `--yes` → auto-approve (`lambda _: True`). Suits CLI scripting where the operator already approved scope.
- `--no-yes` → `click.confirm` on TTY.

The callback is set on the `AgentExecutor`; `_invoke_agent` reads it and forwards it to the constructed agent via `agent_kwargs["interactive"]` and `agent_kwargs["confirm_callback"]`.

## Repro

Configured a stdio MCP server with 78 tools (`CyberStrikeAI cmd/mcp-stdio`) and a template enumerating them in `tools`. Before either patch: agent built with 4 tools, every tool call returned confirmation-error. After both: agent built with 4 native + 78 MCP, `--yes` lets the agent actually call them end-to-end.

(Depends on #339 for the stdio handshake to even complete; without that the discovery hangs before tools land on `system.tool_executor`.)

## Test plan

- [x] `agents ask <id> "..."` against an agent with MCP tools in its template — agent now resolves them and executes (verified with `exec` tool on CyberStrikeAI MCP, ran `nmap` on 127.0.0.1, returned actual scan output).
- [x] `agents ask --no-yes <id> "use shell_exec to run X"` — TTY prompt appears.
- [ ] Adding a regression test for `_invoke_agent` MCP fallback would be welcome — happy to add one if you point me at the preferred mocking pattern.

### DIFF
--- a/src/openjarvis/agents/executor.py
+++ b/src/openjarvis/agents/executor.py
@@ -303,6 +303,21 @@ class AgentExecutor:
                         tool_instances.append(tool)
                     except Exception:
                         logger.warning("Failed to instantiate tool %s", tname)
+
+            # Pull tools already discovered by SystemBuilder (e.g. external MCP
+            # adapters) that aren't in the static ToolRegistry. Without this,
+            # agents declaring MCP-discovered tools in their template would
+            # silently fall back to natives only.
+            if self._system is not None and getattr(self._system, "tool_executor", None) is not None:
+                mcp_pool = getattr(self._system.tool_executor, "_tools", {}) or {}
+                existing = {t.spec.name for t in tool_instances}
+                for tname in tool_names:
+                    if tname in existing:
+                        continue
+                    pooled = mcp_pool.get(tname)
+                    if pooled is not None:
+                        tool_instances.append(pooled)
+
             if tool_instances:
                 logger.info(
                     "Agent %s: resolved %d/%d tools",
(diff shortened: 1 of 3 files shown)


jonsaadfalcon:
Hi @Dilligaf371 — closing this PR because both of your commits landed on `main` as part of PR #367, a consolidation of five contributor PRs grouped to avoid overlapping touches to `mcp/transport.py`, `cli/agent_cmd.py`, and `agents/executor.py`.

Your commits are preserved with your original authorship:
- **`4c37aa83`** — `feat(executor): pull SystemBuilder MCP tools into agent's tool list`
- **`333cb9a3`** — `feat(cli): wire confirm_callback for jarvis agents ask (with --yes/--no-yes)`

No follow-up commits to the logic itself. A small ruff line-wrap on `executor.py:325` ships separately as commit `e087773b` (just satisfies the 88-char E501 limit; no behavior change).

Thanks for closing the loop on `jarvis agents ask` — the system-prompt-describes-tools-but-model-can't-call-them gap and the missing `--yes` for non-interactive use were both quietly hurting MCP UX. `jarvis agents ask --help` now correctly shows the new flag. 🙏

PR #341, opened 2026-05-12
fix(mcp): bump default tool timeout 30s → 600s for MCP-discovered tools

Dilligaf371:
## Summary

`MCPClient.list_tools` constructs `ToolSpec` instances without overriding `timeout_seconds`, so MCP-discovered tools inherit the dataclass default of **30 s**. In practice many MCP servers wrap long-running operations (pentest scanners, build runners, agentic search), so 30 s is often hit before the underlying tool can return.

The symptom is doubly misleading: the failure is reported by the OpenJarvis client side, not by the MCP server, so the cause is invisible from the server logs. The LLM relays the error as natural language ("command execution timed out"), giving no hint that the bound was OpenJarvis-side.

For example: CyberStrikeAI's stdio MCP server allows **30 minutes** per tool via its own `tool_timeout_minutes` config, yet the OpenJarvis client kills the call at 30 s.

## Change

`mcp/client.py` — pass `timeout_seconds=600.0` when building each `ToolSpec` in `list_tools`. Individual MCP servers can still shorten via their own policy.

## Test plan

- [x] Manual: tools that legitimately take ≥30s (full nmap scan, sqlmap session) now complete instead of erroring at 30 s.
- [ ] Adding a unit test on `list_tools` would be welcome — happy to follow your preferred mocking pattern if you point me at one.

(Related to #339 and #340.)

### DIFF
--- a/src/openjarvis/agents/executor.py
+++ b/src/openjarvis/agents/executor.py
@@ -303,6 +303,21 @@ class AgentExecutor:
                         tool_instances.append(tool)
                     except Exception:
                         logger.warning("Failed to instantiate tool %s", tname)
+
+            # Pull tools already discovered by SystemBuilder (e.g. external MCP
+            # adapters) that aren't in the static ToolRegistry. Without this,
+            # agents declaring MCP-discovered tools in their template would
+            # silently fall back to natives only.
+            if self._system is not None and getattr(self._system, "tool_executor", None) is not None:
+                mcp_pool = getattr(self._system.tool_executor, "_tools", {}) or {}
+                existing = {t.spec.name for t in tool_instances}
+                for tname in tool_names:
+                    if tname in existing:
+                        continue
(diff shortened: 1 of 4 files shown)


jonsaadfalcon:
Hi @Dilligaf371 — closing this PR because your fix landed on `main` as part of PR #367, a consolidation of five contributor PRs grouped to avoid overlapping touches to `mcp/transport.py`, `cli/agent_cmd.py`, and `agents/executor.py`.

- **`36482b1a`** — `fix(mcp): bump default tool timeout 30s → 600s for MCP-discovered tools`

No follow-up commits needed.

Thanks for catching this — pentest scanners, build runners, and agentic-search tools really do routinely run past 30s, and "OpenJarvis client timeout reported as natural-language tool error" is exactly the kind of failure mode that makes diagnostics painful. 🙏

PR #355, opened 2026-05-18
feat(memory): persistent memory system for BaseAgent

motors78:
Adds persistent memory system to OpenJarvis BaseAgent.

Features:
- Stores user/assistant interactions
- Builds memory context for prompts
- Prevents duplicate memory entries
- Improves long-term conversation continuity

This is the first step toward long-term agent memory.


fjwood69:
Good first step on persistent memory. Worth knowing about Mori (https://get.moriapp.dev/) as you think about the memory architecture here — it's an MCP-based memory layer that connects via OpenJarvis's native MCP support rather than requiring in-process code changes. Stores memories across sessions and devices, with a dream pipeline that distils interactions into structured, curated memories rather than raw conversation history. The governance angle might be relevant for OpenJarvis too — agent-written memories go through an intake pipeline before reaching the shared store.


Happy to discuss integration if useful - fw

ElliotSlusky:
Closing after maintainer review. The branch contains unresolved conflict markers and an obsolete unlocked/non-atomic persistence implementation that has been superseded by the current memory subsystem. Closing.

Issue #356, opened 2026-05-18
No Mac Download


micahzakashansky-design:
The Mac Silicon Download link is invalid:
https://github.com/open-jarvis/OpenJarvis/blob/b863cbb07b9837fa6358d6e6d7cc6c27d80ee097/docs/downloads.md?plain=1#L28

Rari-GitH:
Still inactive.  Where is MAC download

micahzakashansky-design:
I found a working download at: https://github.com/open-jarvis/OpenJarvis/releases
https://release-assets.githubusercontent.com/github-production-release-asset/1158198234/c28a0411-3b04-4c2b-93a7-93a310b69fd7?sp=r&sv=2018-11-09&sr=b&spr=https&se=2026-05-19T00%3A32%3A04Z&rscd=attachment%3B+filename%3DOpenJarvis_0.1.0_universal.dmg&rsct=application%2Foctet-stream&skoid=96c2d410-5711-43a1-aedd-ab1947aa7ab0&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skt=2026-05-18T23%3A31%3A23Z&ske=2026-05-19T00%3A32%3A04Z&sks=b&skv=2018-11-09&sig=FH2X6z21Xb4w0mjyBBPHUalFDIG1GRv0oKTI5%2Fsa6EM%3D&jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmVsZWFzZS1hc3NldHMuZ2l0aHVidXNlcmNvbnRlbnQuY29tIiwia2V5Ijoia2V5MSIsImV4cCI6MTc3OTE1MDMzMiwibmJmIjoxNzc5MTQ4NTMyLCJwYXRoIjoicmVsZWFzZWFzc2V0cHJvZHVjdGlvbi5ibG9iLmNvcmUud2luZG93cy5uZXQifQ.OP4II1QBO4Kx5JD_jVTZF4w5wgLORnz-WaJcVXv96pA&response-content-disposition=attachment%3B%20filename%3DOpenJarvis_0.1.0_universal.dmg&response-content-type=application%2Foctet-stream
but now that is also invalid, sorry.

micahzakashansky-design:
or wait maybe that works, for some reason it didn't work a couple of seconds ago but it does now.

robbym-dev:
[...]

PR #374, opened 2026-05-22
Update downloads.md

micahzakashansky-design:
Purpose:
Fixing the Mac download link

Outcome:
Used and performed successfully


robbym-dev:
Thanks for this! The Mac download link was already fixed on `main` by #412 — `docs/downloads.md` now points at the stable `desktop-v1.0.2` universal `.dmg`, which resolves (HTTP 200). This PR has also gone stale: it now conflicts with `main`, and its target URL is an expired short-lived signed `release-assets` URL (expiry `2026-05-19`) wrapped in malformed nested-markdown — so merging it would actually re-break the Mac download. Closing as superseded by #412. See #356.

PR #381, opened 2026-05-22
fix: improve Windows compatibility in quickstart script

sanjayravit:
- Handle 'python' as fallback for 'python3'\n- Add Windows support for opening browser via 'cmd /c start'

## What does this PR do?

This PR improves Windows compatibility in the quickstart script by making two key changes:

Python command fallback: Instead of only checking for python3, the script now falls back to python if python3 is not available. This is important because Windows users often only have python in their PATH.

Windows browser opening: Adds support for opening URLs on Windows systems using cmd /c start, which is the Windows equivalent of open (macOS) and xdg-open (Linux).

## Checklist

- [ ] Linter passes (`uv run ruff check src/ tests/`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ jarvis
 
 The Rust extension and bigger models continue downloading in the background while you chat. Run `jarvis doctor` to see status.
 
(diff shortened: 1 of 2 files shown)


sanjayravit:
for creating new one

PR #383, opened 2026-05-22
fix(server): preserve tool_calls when rebuilding streaming-chat history

### LINKED ISSUE #382: Streaming managed-agent endpoint drops tool_calls when rebuilding history → multi-turn regression to fabricated tool output
## Summary

`_stream_managed_agent` in `src/openjarvis/server/agent_manager_routes.py` rebuilds the per-turn conversation history by appending each prior assistant message as `Message(role=ASSISTANT, content=m["content"])` and discarding the stored `tool_calls` field entirely. The model never sees its own prior tool-use pattern and regresses to fabricating tool output on turn 2+.

## Repro

1. Create a managed agent with at least one tool wired:

```bash
curl -X POST http://localhost:8000/v1/managed-agents \
  -H 'content-type: application/json' \
  -d '{"name":"ToolTest","agent_type":"monitor_operative",
       "config":{"model":"<your-model>","tools":["shell_exec"],
                 "system_prompt":"Use tools when asked. Do not narrate."}}'
```

2. Turn 1 — works correctly:

```bash
curl -N -X POST http://localhost:8000/v1/managed-agents/<id>/messages \
  -H 'content-type: application/json' \
  -d '{"content":"Use shell_exec to run pwd.","mode":"immediate","stream":true}'
```

Observed: `event: tool_call_start` for `shell_exec`, then text. ✓

3. Turn 2 — regresses:

```bash
curl -N -X POST http://localhost:8000/v1/managed-agents/<id>/messages \
  -H 'content-type: application/json' \
  -d '{"content":"Now run shell_exec with ls.","mode":"immediate","stream":true}'
```

Observed: **no `tool_call_start` event.** The model fabricates a plausible `ls` output and never invokes the tool.

The stored tool_calls JSON for the turn-1 row is intact (`agent_messages.tool_calls IS NOT NULL`); only the *replay* path drops it.

## Root cause

In `src/openjarvis/server/agent_manager_routes.py` (current main, around 660–668):

```python
history = manager.list_messages(agent_id, limit=50)
for m in reversed(history):
    if m["id"] == message_id:
        continue
    if m["direction"] == "user_to_agent":
        llm_messages.append(Message(role=Role.USER, content=m["content"]))
    elif m["direction"] == "agent_to_user":
        llm_messages.append(Message(role=Role.ASSISTANT, content=m["content"]))
```

`m["tool_calls"]` — already parsed from JSON by `AgentManager.list_messages` — is dropped. OpenAI chat-completions expects each assistant turn with `tool_calls` to be followed by one `role: tool` message per result; we send neither. To the model, prior turns look like "user asked → assistant wrote prose summary," so the pattern it learns is to summarize without calling.

`store_agent_response` collapses persisted tool_calls to `{tool, arguments, result, success, latency}` for UI replay — the original OpenAI `tool_call_id` isn't kept. The fix synthesizes stable IDs.

## Proposed fix

```python
elif m["direction"] == "agent_to_user":
    stored_tcs = m.get("tool_calls") or []
    if not stored_tcs:
        llm_messages.append(Message(role=Role.ASSISTANT, content=m["content"]))
        continue
    synth_ids = [f"call_{m['id']}_{i}" for i in range(len(stored_tcs))]
    llm_messages.append(Message(
        role=Role.ASSISTANT,
        content=m["content"] or None,
        tool_calls=[
            MsgToolCall(id=synth_ids[i],
                        name=tc.get("tool",""),
                        arguments=tc.get("arguments",""))
            for i, tc in enumerate(stored_tcs)
        ],
    ))
    for i, tc in enumerate(stored_tcs):
        llm_messages.append(Message(
            role=Role.TOOL,
            tool_call_id=synth_ids[i],
            content=tc.get("result") or "",
        ))
```

PR with this fix to follow.

## Verified

With the proposed fix applied locally, turn 2's `event: tool_call_start` returns. The agent continues using its registered tool surface across turns instead of degrading to fabrication.

## Environment

- OpenJarvis installed from source at commit 07b6cba (current `main`)
- Engine: `mlx` against `mlx_lm.server` with qwen3.6-35b-a3b
- Behavior should reproduce against any backend whose chat template requires the assistant-`tool_calls` / `role: tool` pairing to signal in-context tool use.

## Why it matters

For native chat apps consuming `/v1/managed-agents/{id}/messages` (the natural integration point for client UIs), multi-turn tool use is the daily case. Without this, an agent's registered tools effectively go invisible after turn 1 — present in config, ignored in practice.

gilbert-barajas:
## Summary

`_stream_managed_agent` rebuilt prior assistant turns as plain content, dropping the stored `tool_calls`. After turn 1, the model lost its own tool-use pattern and regressed to fabricating tool output instead of invoking tools — even though the tools stayed wired in config and ToolRegistry the whole time.

This restores the proper OpenAI shape on replay — assistant turn carrying `tool_calls` followed by one `role: tool` message per result.

## Why ids are synthesized

`store_agent_response` already collapses tool_calls to `{tool, arguments, result, success, latency}` for UI replay; the original OpenAI `tool_call_id` isn't persisted. Models only need the assistant `tool_calls` and matching `role: tool` messages to reference the same string, so a stable per-row synthesis (`call_{message_id}_{i}`) is sufficient without a schema migration.

## Verified

Repro and verification in #382 (linked). Reproduces against `mlx` engine + qwen3.6-35b on default `monitor_operative` agent. With this patch:

- Turn 1: `event: tool_call_start` for the requested tool. ✓ (unchanged)
- Turn 2: `event: tool_call_start` returns. ✓ (was: no event; fabricated output)

Test plan:

- [x] Manual two-turn repro from #382 (turn 1 + turn 2 both invoke tools)
- [x] Pre-commit `ruff` + `ruff-format` pass
- [ ] CI

Closes #382.

### DIFF
--- a/src/openjarvis/server/agent_manager_routes.py
+++ b/src/openjarvis/server/agent_manager_routes.py
@@ -632,6 +632,7 @@ async def _stream_managed_agent(
     import uuid
 
     from openjarvis.core.types import Message, Role
+    from openjarvis.core.types import ToolCall as MsgToolCall
 
     agent_id = agent_record["id"]
     config = agent_record.get("config", {})
@@ -657,7 +658,13 @@ async def _stream_managed_agent(
         if app_state is not None and dr_tools:
             app_state._dr_tools = dr_tools
 
-    # Load prior conversation context (DESC order, reverse for chronological)
+    # Load prior conversation context (DESC order, reverse for chronological).
+    # Assistant turns with stored tool_calls are replayed as the proper
+    # OpenAI shape — assistant message with tool_calls followed by one
+    # tool-role message per result — so the model retains its prior
+    # tool-use pattern across turns. Without this, an agent that used
+    # tools on turn 1 sees only a plain prose summary in history and
+    # regresses to fabricating tool output on turn 2+.
     history = manager.list_messages(agent_id, limit=50)
     for m in reversed(history):
         if m["id"] == message_id:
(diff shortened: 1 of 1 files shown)


gilbert-barajas:
Closing — landed upstream in #418, which consolidates this with #386 and #395 (all three touched `_stream_managed_agent` and would have conflicted as separate PRs). Good call on the consolidation + the added `test_managed_agent_streaming.py` coverage. Thanks!

PR #387, opened 2026-05-22
feat(server): expose sampler params (repetition_penalty / top_p / top_k / min_p / *_penalty) for managed agents

### LINKED ISSUE #386: Managed-agent streaming doesn't forward sampler params (repetition_penalty / top_p / top_k / min_p / *_penalty) to the upstream engine
## Problem

`_stream_managed_agent` in `src/openjarvis/server/agent_manager_routes.py` reads only `temperature` and `max_tokens` from a managed agent's `config_json` and forwards them to `engine.stream_full(...)`. There's no way to set `repetition_penalty`, `top_p`, `top_k`, `min_p`, `frequency_penalty`, or `presence_penalty` per agent, even though the OpenAI-compat engine path already passes `**kwargs` through to the upstream `/v1/chat/completions` payload — so the plumbing exists; it just isn't read from config.

The practical consequence: any locally-hosted model with non-default sampler needs (most of them, in practice) can't be properly tuned via a managed agent. Most painfully, **no `repetition_penalty` ≈ degenerate repetition loops on common open-weights models.**

## Concrete repro

A managed agent backed by `qwen3.6-35b-a3b` (any vendor's MLX quant — Unsloth or Huihui-abliterated both reproduce it) served by `mlx_lm.server` with no sampler config will, with non-trivial frequency, emit the same paragraph N times **in a single completion** before EOS. Observed in one of my own sessions: 32× repetition of the same ~150-token paragraph in one streamed response, ~240 KB of SSE before `finish_reason: "stop"` finally fired.

Adding `repetition_penalty: 1.1` to the upstream completion call closes this completely, but with the current code I had to set it at the serving-layer level (`--default-repetition-penalty 1.1` on Rapid-MLX) rather than per-agent in OpenJarvis. That works for me but doesn't scale across agents that share the same engine.

## Where it would slot in

```python
# agent_manager_routes.py:640-641 (today)
temperature = config.get("temperature", 0.7)
max_tokens  = config.get("max_tokens", 1024)
```

```python
# agent_manager_routes.py:958-961 + 1053-1055 (today)
stream_kwargs: Dict[str, Any] = {}
resolved_tools = _resolve_tool_specs(config.get("tools"))
if resolved_tools:
    stream_kwargs["tools"] = resolved_tools
...
async for chunk in engine.stream_full(
    messages_for_llm,
    model=model,
    temperature=temperature,
    max_tokens=max_tokens,
    **stream_kwargs,
):
```

`_OpenAICompatibleEngine.stream_full` already builds the payload as `{..., **kwargs}` and forwards to the upstream `chat/completions`, so kwargs *do* land in the wire payload. The gap is just reading them from `config` and merging into `stream_kwargs`.

## Proposed shape

Read a known set of sampler params from `config_json` and merge into `stream_kwargs` if present. Skip if absent (full backward compat — current agents keep their current behavior):

```python
SAMPLER_PARAMS = (
    "repetition_penalty", "top_p", "top_k", "min_p",
    "frequency_penalty", "presence_penalty",
)
for k in SAMPLER_PARAMS:
    if k in config:
        stream_kwargs[k] = config[k]
```

Whether `top_p` / `top_k` etc are honored by the upstream is engine-specific — but that's already how `tools` works (forwarded blindly; some engines accept, others 400). The OpenAI-compat path already swallows unknown kwargs gracefully via the existing `try / payload.pop` pattern at `engine/_openai_compat.py:63-64`.

## Why issue-first

Touches the agent `config_json` schema (de facto, even if the column is `TEXT`), so I wanted to confirm:

1. Are these the right param names? (OpenAI uses `frequency_penalty` / `presence_penalty`; vLLM and Rapid-MLX use `repetition_penalty`. I'd add both and let the engine pick.)
2. Should there be a global default in app config that gets merged in, à la `temperature: 0.7`?
3. Anything in the existing managed-agent CLI / UI for creation that I should update in the same PR?

Happy to send the PR once we agree on the surface area.

## Environment

- OpenJarvis @ `main` (post #371 cleanup)
- macOS 14.x / Apple Silicon (M5 Max)
- Engine: locally `mlx_lm.server` and `Rapid-MLX serve` both reproduce

gilbert-barajas:
Closes #386.

## What this changes

`_stream_managed_agent` now reads six sampler parameters from a managed agent's `config_json` and merges them into the kwargs passed to `engine.stream_full(...)`:

- `repetition_penalty`
- `top_p`
- `top_k`
- `min_p`
- `frequency_penalty`
- `presence_penalty`

Skipped silently when absent — existing agents see no behavior change.

## Why

`_OpenAICompatibleEngine.stream_full` already builds the upstream payload as `{..., **kwargs}` and forwards to `/v1/chat/completions`, so the plumbing already exists. The gap was just *reading* these from `config_json` and merging into `stream_kwargs`. This PR closes that gap.

Concrete motivating bug from my own session: managed agent backed by `qwen3.6-35b-a3b` (MLX 8-bit, Unsloth's clean build or any vendor's) emitted **the same paragraph 32× in one completion** — ~240 KB of SSE before `finish_reason: "stop"` finally fired. Setting `repetition_penalty: 1.1` on the agent record now closes that without per-server-process knobs.

See #386 for the full repro + reasoning.

## What this does NOT change

- No schema migration. `config_json` is `TEXT`; the new keys are read opportunistically.
- No new defaults injected. Agents without these keys behave exactly as before.
- No engine-side changes. The OpenAI-compat engine's existing `**kwargs` passthrough + 400-error tolerance (`engine/_openai_compat.py:63-64`) handles this for free.
- Not touched: `DeepResearchAgent` path (separate branch at `agent_manager_routes.py:679`+). I scoped this PR to the main streaming branch only; happy to extend in a follow-up if you'd like.

## Tests

Two new tests added to `tests/server/test_agent_manager_routes.py::TestAgentManagerStreaming`:

1. `test_send_message_stream_forwards_sampler_params` — agent configured with all six sampler params; mock engine captures `stream_full` kwargs; asserts all six land, and that unrelated config keys (`custom_meta`, `owner_email`) do NOT leak.
2. `test_send_message_stream_omits_unset_sampler_params` — backward-compat: an agent with no sampler config produces a kwargs dict that doesn't include any of the six keys.

Both pass. Full test file (30 tests including the existing streaming suite): `30 passed`.

```
$ pytest tests/server/test_agent_manager_routes.py -q
..............................                                           [100%]
30 passed, 1 warning in 0.80s
```

## Diff size

`+128 / -0`, two files (`agent_manager_routes.py` +15, the rest is tests).

### DIFF
--- a/src/openjarvis/server/agent_manager_routes.py
+++ b/src/openjarvis/server/agent_manager_routes.py
@@ -960,6 +960,25 @@ async def _stream_managed_agent(
     if resolved_tools:
         stream_kwargs["tools"] = resolved_tools
 
+    # Per-agent sampler params. The OpenAI-compat engine already forwards
+    # unknown kwargs into the upstream /v1/chat/completions payload, so the
+    # plumbing already works — this just lets the agent's stored config
+    # drive it. Skipped if absent (full backward compat with existing agents).
+    # Names cover both OpenAI-style (frequency/presence_penalty, top_p) and
+    # llama.cpp / vLLM / mlx-style (repetition_penalty, top_k, min_p); the
+    # upstream engine accepts what it accepts and the _OpenAICompatibleEngine
+    # already swallows 400s for unknown payload keys (engine/_openai_compat.py).
+    for _sampler_key in (
+        "repetition_penalty",
+        "top_p",
+        "top_k",
+        "min_p",
+        "frequency_penalty",
+        "presence_penalty",
+    ):
+        if _sampler_key in config:
+            stream_kwargs[_sampler_key] = config[_sampler_key]
+
     # Discover MCP tools and merge into stream_kwargs
     mcp_adapters: Dict[str, Any] = {}
     if app_state is not None:
--- a/tests/server/test_agent_manager_routes.py
+++ b/tests/server/test_agent_manager_routes.py
@@ -385,6 +385,115 @@ class TestAgentManagerStreaming:
         # Last chunk should have finish_reason="stop"
         assert chunks[-1]["choices"][0]["finish_reason"] == "stop"
 
+    def test_send_message_stream_forwards_sampler_params(self, manager):
+        """Per-agent sampler params (repetition_penalty, top_p, top_k, min_p,
+        frequency_penalty, presence_penalty) configured on the agent flow
+        through to engine.stream_full(**kwargs). Other unrelated config keys
+        do NOT leak into kwargs."""
+        from openjarvis.engine._stubs import StreamChunk
+
+        captured_kwargs: dict = {}
+
+        async def _stream_full(messages, *, model, **kwargs):
+            captured_kwargs.update(kwargs)
+            yield StreamChunk(content="ok")
+            yield StreamChunk(finish_reason="stop")
+
+        engine = MagicMock()
+        engine.engine_id = "mock"
+        engine._model = "test-model"
+        engine.stream_full = _stream_full
+
+        from fastapi import FastAPI
+        from fastapi.testclient import TestClient as TC
+
+        from openjarvis.server.agent_manager_routes import create_agent_manager_router
+
+        app = FastAPI()
+        app.state.engine = engine
+        app.state.bus = None
+        routers = create_agent_manager_router(manager)
+        for r in routers:
+            app.include_router(r)
+        client = TC(app)
+
+        agent = manager.create_agent(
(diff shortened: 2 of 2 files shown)


gilbert-barajas:
Closing — landed upstream in #418, consolidated with #382 and #395 since all three touched the same function in `agent_manager_routes.py`. Thanks!

PR #451, opened 2026-05-30
fix(security): detect Rust at import time + SSRF Python fallback (#225)

therahul-yo:
Fixes #225.

## Problem

`src/openjarvis/_rust_bridge.py` hardcoded the exported availability flag:

```python
RUST_AVAILABLE: bool = True
```

This value was unconditional, so it never reflected whether the compiled `openjarvis_rust` extension was actually built. Two concrete consequences when the extension is **not** compiled (fresh clone without `maturin develop`, CI without a Rust toolchain, a pip install that didn't build the backend):

1. **`tools/browser.py` silently disabled SSRF protection.** The SSRF check was wrapped in `try: ... except ImportError: pass`. With no compiled backend, `check_ssrf()` raised `ImportError` (from `get_rust_module()`), which was swallowed — so the browser tool would navigate to internal/cloud-metadata endpoints (e.g. `169.254.169.254`) **with no SSRF check at all**.
2. **`tools/http_request.py` and `tools/web_search.py` crashed.** They call `check_ssrf()` unguarded, so the same `ImportError` took down those tools instead of degrading gracefully.

A complete pure-Python SSRF implementation (`_check_ssrf_python`) already existed in `ssrf.py` but was labelled "kept for reference only" and was never reachable — even though the test module already describes it as "the pure-Python implementation".

## Fix

- **`_rust_bridge.py`** — compute `RUST_AVAILABLE` dynamically by probing the extension once at import time, so the exported flag is honest. `get_rust_module()` still hard-raises `ImportError` for the modules where Rust is genuinely mandatory.
- **`security/ssrf.py`** — `check_ssrf()` now uses the Rust backend when available and falls back to `_check_ssrf_python()` otherwise, instead of raising. The SSRF guard is security-critical and must never be silently skipped or crash merely because Rust wasn't built.
- **`tools/browser.py`** — removed the `except ImportError: pass` that disabled SSRF protection; the check now always runs (Rust or Python).
- **`tests/security/test_ssrf.py`** — added `TestCheckSsrfPythonFallback` covering the fallback path (cloud-metadata IP, private IP via DNS, and a public URL) with `RUST_AVAILABLE` patched to `False`.

## Verification

- `ruff check` passes on all changed files.
- New and existing SSRF logic verified by loading `ssrf.py` with a stubbed (unavailable) Rust bridge: metadata + private IPs are blocked via the Python path, public URLs pass, and no `ImportError` escapes.

### DIFF
--- a/src/openjarvis/_rust_bridge.py
+++ b/src/openjarvis/_rust_bridge.py
@@ -32,7 +32,21 @@ def get_rust_module() -> _types.ModuleType:
     return openjarvis_rust
 
 
-RUST_AVAILABLE: bool = True
+def _detect_rust() -> bool:
+    """Return ``True`` if the compiled ``openjarvis_rust`` extension is importable.
+
+    Computed once at import time. Modules with a Python fallback (e.g.
+    ``security.ssrf``) consult this flag instead of hardcoding availability,
+    so the fallback is actually reachable when the extension was not built.
+    """
+    try:
+        get_rust_module()
+    except ImportError:
+        return False
+    return True
+
+
+RUST_AVAILABLE: bool = _detect_rust()
 
 
 # ---------------------------------------------------------------------------
--- a/src/openjarvis/security/ssrf.py
+++ b/src/openjarvis/security/ssrf.py
@@ -70,15 +70,23 @@ def is_private_ip(ip_str: str) -> bool:
 
 
 def check_ssrf(url: str) -> Optional[str]:
-    """Check a URL for SSRF vulnerabilities — always via Rust backend."""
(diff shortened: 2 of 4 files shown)


jonsaadfalcon:
Superseded by #467, which cherry-picks your commit verbatim (authorship preserved) and adds two small follow-ups: a backend-agnostic test_no_hostname assertion and an updated browser SSRF test (the old test_execute_ssrf_module_missing asserted the skip-on-ImportError behavior this fix deliberately removes). Full credit to @therahul-yo for the fix. Thank you!

Issue #476, opened 2026-06-02
Installer scrip hard-coded python 3.11 but the project allows newer: `requires-python = ">=3.10, <3.14"`


senki:
### Description

See :

https://github.com/open-jarvis/OpenJarvis/blob/58f7f717aef3aac7139ece526c3a8f438df89421/uv.lock#L3

and:

https://github.com/open-jarvis/OpenJarvis/blob/58f7f717aef3aac7139ece526c3a8f438df89421/scripts/install/install.sh#L427-L448

### Steps to Reproduce

1 install OpenJarvis

### Expected Behavior

Use a newer Python interpreter.

Optionally, sync with the project requirements to avoid manual updates in the future.

### Actual Behavior

OpenJarvis downloads Python 3.11 even in the presence of 3.12 or 3.13, which are actually compatible with the project. 

### Operating System

macOS

### Python Version

3.11

### Hardware

Apple Silicon

### Engine

None

### Logs / Traceback

```shell

```

jonsaadfalcon:
Thanks @senki — the lockfile vs install.sh asymmetry is real, you're right. `pyproject.toml` declares `requires-python = ">=3.10,<3.14"` (the upper bound is the #432 numpy cp314 gate), but the installer hardcodes 3.11.

This pairs with #484 (Sanjay's robustness fix for Python-less hosts). I'm rolling both concerns into a consolidated install.sh change that:

1. Prefers the newest Python already on the host in the `>=3.10,<3.14` range
2. Falls back to `uv python install` only when none qualify (still picks the newest allowed version, currently 3.13)
3. Fixes the analytics/state-file shims to not crash when no system Python exists (orthogonal to #444's `create_venv` bootstrap — the shims run earlier in the pipeline)

The hardcoded `3.11` was a conservative pick at the time (3.11 had broadest wheel coverage across our optional extras); now that 3.13's wheel ecosystem has caught up there's no reason not to use the newest in range. PR coming.

PR #484, opened 2026-06-03
fix(install): gracefully handle host environments without Python

sanjayravit:
## What does this PR do?
This PR makes the [install.sh](cci:7://file:///e:/jarvis/scripts/install/install.sh:0:0-0:0) script more robust by gracefully handling host environments where `python3` or `python` is not yet available on the PATH.

Previously, the script would crash early at the analytics beacon or state-marking stages if no host Python was found. This change adds a fallback to `true` for the `PY_CMD` variable, allowing the script to proceed through the [install_uv](cci:1://file:///e:/jarvis/scripts/install/install.sh:404:0-413:1) and [create_venv](cci:1://file:///e:/jarvis/scripts/install/install.sh:428:0-449:1) steps where a managed Python version is eventually installed.

## Why is this important?
- Enables a smoother "cold start" experience on fresh systems.
- Prevents unnecessary script failures in minimalist environments.
- Maintainers requested smaller, focused PRs instead of large rewrites.

## How was this tested?
- Verified the logic on a system with Python installed (normal operation).
- Manually tested by temporarily mocking a missing Python to ensure the script continues without error.

## Checklist

- [x] Follows registry pattern (if adding new component)

### DIFF
--- a/scripts/install/install.sh
+++ b/scripts/install/install.sh
@@ -202,12 +202,14 @@ need curl
 # on minimal distros that only ship the unversioned name. This is used by
 # the analytics beacon and state-file helpers below — uv installs its own
 # interpreter into the venv later, so this is only for the bootstrap shims.
-PY_CMD="python3"
-if ! command -v python3 >/dev/null 2>&1; then
-    if command -v python >/dev/null 2>&1; then
-        PY_CMD="python"
-    fi
+PY_CMD=""
+if command -v python3 >/dev/null 2>&1; then
+    PY_CMD="python3"
+elif command -v python >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


sanjayravit:
@jonsaadfalcon

robbym-dev:
Superseded by #492, now merged. #492 incorporates this PR's core idea — letting `install.sh` proceed gracefully on hosts where `python3`/`python` isn't yet on PATH — and additionally resolves the hardcoded-`3.11` version issue (#476) by reading the `requires-python` range. It passed CI across all 11 platform/lint/bats checks. Thanks for the original fix; its approach is preserved in the merged version. Closing as superseded.

PR #492, opened 2026-06-03
fix(install): use requires-python range; bootstrap shims are Python-free (#476, #484)

### LINKED ISSUE #476: Installer scrip hard-coded python 3.11 but the project allows newer: `requires-python = ">=3.10, <3.14"`

jonsaadfalcon:
## Summary

Consolidates two overlapping reports about \`scripts/install/install.sh\`:

- **#476** (@senki): installer hardcoded \`--python 3.11\` but pyproject declares \`requires-python = \">=3.10,<3.14\"\`. Use the newest version in range.
- **#484** (@sanjayravit): installer crashed on hosts with no \`python3\`/\`python\` on PATH because the bootstrap shims (mark_done, beacon, get_anon_id) all called inline Python heredocs.

Both close with five surgical edits to install.sh — all behavior-preserving for the existing happy path, proven by the bats tests.

## Edits

1. **get_anon_id** — POSIX UUID v4 via \`/dev/urandom + od\` + bash substring (replaces \`python3 -c 'import uuid; ...'\`).
2. **beacon** — curl + shell-built JSON payload (replaces 40-line Python heredoc). Inputs are from controlled sources so no general-purpose JSON escaping needed; \`|| true\` is load-bearing under the ERR trap.
3. **mark_done** — awk regenerator (replaces \`python3 -c 'json.load+update+dump'\`). Idempotent: early-return if key already marked. The wsl flag is always rewritten last so a later \`FORCE_WSL=1\` re-run correctly updates it.
4. **parse_requires_python** — new helper. Reads pyproject's requires-python; handles inclusive (\`<=3.13\`) and exclusive (\`<3.14\`) upper bounds correctly; falls back to 3.11 if it can't parse.
5. **create_venv** — call \`parse_requires_python\` instead of hardcoding 3.11. The existing uv-managed-Python fallback (from #444) still kicks in if the host doesn't have the target version installed.

## Adversarial review

Found two real bugs in the first draft, both fixed pre-commit:

- **HIGH**: \`parse_requires_python\`'s exclusive-bound regex would also match \`<=3.13\` (inclusive) and subtract 1, producing 3.12. Fixed by checking the inclusive form first.
- **MEDIUM**: the new \"no Python on PATH\" bats test silently skipped symlinking \`pgrep\` on hosts where it's not at /usr/bin or /bin (BusyBox/Alpine). Added a \"no matching process\" stub fallback.

Other adversarial concerns surveyed (10 total) — all either non-issues or out-of-scope (e.g. embedded quotes in step names — not exploitable with current step vocabulary).

## Test coverage

29/29 install bats tests pass, including 3 new ones:

- **\`install succeeds with no system Python on PATH (#484)\`** — builds a PATH without python3/python; asserts the install completes and state file is written.
- **\`mark_done is idempotent — second mark of same key doesn't duplicate\`** — re-runs install; asserts \`install_uv\` appears exactly once.
- **\`create_venv picks newest in requires-python range, not hardcoded 3.11 (#476)\`** — asserts uv was called with \`--python 3.13\`.

Git stub now sets \`requires-python = \">=3.10,<3.14\"\` in its fake pyproject so the test exercises real parsing.

## Closes / supersedes

- Closes #476
- Closes #484 (the existing PR — superseded with credit to @sanjayravit who motivated this consolidation)

### DIFF
--- a/scripts/install/install.sh
+++ b/scripts/install/install.sh
@@ -262,15 +262,19 @@ detect_arch() {
 }
 
 get_anon_id() {
+    # POSIX shell UUID v4 — no Python required so this works on hosts
+    # where python3/python aren't on PATH yet (#484). The script must
+    # not crash on those hosts; analytics are best-effort.
     if [[ -f "$ANON_ID_FILE" ]]; then
         cat "$ANON_ID_FILE"
         return
     fi
-    local new_id
-    new_id="$("$PY_CMD" -c 'import uuid; print(uuid.uuid4())' 2>/dev/null || echo "")"
-    if [[ -z "$new_id" ]]; then
-        return
+    local raw new_id
+    raw="$(od -An -tx1 -N16 /dev/urandom 2>/dev/null | tr -d ' \n')" || true
+    if [[ ${#raw} -ne 32 ]]; then
+        return 0  # /dev/urandom unreadable — skip analytics silently
     fi
+    new_id="${raw:0:8}-${raw:8:4}-${raw:12:4}-${raw:16:4}-${raw:20:12}"
     echo "$new_id" > "$ANON_ID_FILE"
     echo "$new_id"
 }
@@ -291,6 +295,13 @@ stage_label() {
 
 beacon() {
     # Args: event_name stage_label elapsed_ms exit_code
+    #
+    # No Python required (#484) — uses curl + shell-built JSON. All
+    # inputs are from controlled sources: $event is a fixed-vocabulary
+    # string (install_started / install_stage_completed / install_failed
+    # / install_completed), $stage comes from stage_label(), the numeric
+    # args are validated by the arithmetic that produced them, and
+    # $anon_id is a fresh UUID. No general-purpose JSON escaping needed.
     local event="$1"
     local stage="${2:-}"
     local elapsed_ms="${3:-0}"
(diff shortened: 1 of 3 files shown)


robbym-dev (approved):
Approved: replaces fragile inline-Python heredocs in install.sh with POSIX shell/awk/curl and reads the requires-python range instead of hardcoding 3.11. CI green across all 11 platform/lint/bats checks. Supersedes #484, closes #476.

Issue #475, opened 2026-06-02
[RFC] Measuring scaffolding-dependence: a harnessability taxonomy + ablation for spec search & self-improving operators


gilbert-barajas:
*(Discussion issue — proposing an evaluation methodology + framework that feeds the Learning roadmap. Research-flavored; offered as a collaboration, n=1 prelim flagged honestly.)*

## Motivation

Three roadmap/feature lines all optimize a harness from traces but lack a shared *measurement* of how much harness a given model actually needs:
- **LLM-guided spec search** (user-guide) — frontier model analyzes traces, proposes config/spec improvements.
- **Self-improving operators** (Research-Stage) — operators tune their own prompts + routing from trace feedback.
- **Learning** (architecture/learning) + the evals framework, which already treats energy / FLOPs / latency / dollar cost as first-class.

To *optimize* a spec and *know it helped*, you need to answer, per capability: **how much does this local model supply on its own vs. need scaffolded — and which scaffolding actually recovers the capability?** We propose adding **scaffolding-dependence** as a first-class eval axis, with a taxonomy and a controlled instrument.

## Proposal

**(a) The harnessability ladder** — a taxonomy of the properties a task needs, ranked by how deterministically each can be supplied or checked:
1. *oracle-able* (a deterministic check exists — compile/test; scope gates; **domain-semantic correctness via metamorphic property tests**),
2. *retrievable* (surface the right context — signature/spec retrieval),
3. *substrate-encodable* (move the knowledge into what the model reads — e.g. metric semantics in docstrings),
4. *critique-able* (no clean oracle, adversarial review catches most),
5. *residue* (open-ended judgment — the genuine frontier premium).

This tells spec-search *where* a property can be moved (promote runtime judgment → a compile-time oracle).

**(b) The spec-tightness × model-role ablation** — hold model + task fixed; vary **spec detail** (full vs intent-only) and **model role** (coder vs reasoner). The delta quantifies, per property, how much the model supplies itself vs needs scaffolded — a reproducible signal the spec-search loop can optimize against and self-improving operators can use as a reward.

[...]

PR #493, opened 2026-06-03
feat(prompt): per-invocation persona scope (#380)

### LINKED ISSUE #380: Persona files (SOUL/MEMORY/USER) are loaded globally for every `jarvis ask`, with no per-invocation opt-out — breaks multi-agent setups
## Summary

This is an observation + design question, not a confident bug report.

`SystemPromptBuilder` ([src/openjarvis/prompt/builder.py:48](https://github.com/open-jarvis/OpenJarvis/blob/main/src/openjarvis/prompt/builder.py#L48)) reads `MemoryFilesConfig.soul_path` / `memory_path` / `user_path` and unconditionally prepends them to the system prompt. Those paths come from a single global `MemoryFilesConfig` (defaults: `~/.openjarvis/SOUL.md`, `~/.openjarvis/MEMORY.md`, `~/.openjarvis/USER.md`) and I can't find a per-invocation flag on `jarvis ask` to swap them out or opt out.

In a multi-agent setup where the user runs more than one model through OpenJarvis (e.g. one model for the daily-driver persona, a second smaller/faster model alongside for quick rewrites, a third on a different engine for a different role), every invocation inherits the *daily-driver* SOUL.md. The second and third model role-play as the daily-driver persona regardless of which engine answered.

This pairs with #376 (persistent agents bypass persona files entirely) as the **inverse** problem — together they suggest persona scope is currently neither fully on (broken in persistent-agent path) nor scopable (broken in one-shot path):

| Path | SOUL.md loaded? | Per-invocation override? |
| --- | --- | --- |
| `jarvis ask` (one-shot) | Yes, always | No |
| `jarvis agents ask` (persistent, #376) | No (bug) | N/A |

## Repro

Default `~/.openjarvis/SOUL.md` describes persona \"A\". From the CLI, invoke ask against a deliberately different model/engine that you intend to behave as a *different* persona — the response still adopts persona A because SOUL.md was prepended.

There is no `--persona`, `--no-soul`, `--memory-files`, or similar flag on `ask.py` that I could find (commit `07b6cba`).

## Design question

Is this by design (one user, one persona — the OS-level singular identity model) and a feature request to add per-invocation persona selection? Or is the intent that engine/model selection *should* be able to imply persona scope and the conflation is unintended?

If a per-invocation override is the desired direction, a few shapes:
- `--persona NAME` resolving to `~/.openjarvis/personas/NAME/SOUL.md` etc.
- Engine-config override block in `~/.openjarvis/config.yaml` letting an engine declare its own `memory_files`.
- A `--no-persona` opt-out for callers that want a clean model.

Happy to take direction here before writing any code — wanted to surface the observation now since #376 fixes one half of persona scoping and this is the other half.

## Environment

- Discovered while wiring multiple chat participants on different engines through Velum (third-party SwiftUI client). One engine = one persona was the assumption; the leak surfaced when a second engine adopted the first engine's identity unprompted.
- `openjarvis` installed from source at `main` HEAD `07b6cba`.

gilbert-barajas:
## Summary

Per-invocation persona selection — one OpenJarvis install can answer as different personas without swapping global config.

- `--persona NAME` / `--persona none` on `jarvis ask`
- `[memory_files].persona_name` config field
- Named personas resolve to `~/.openjarvis/personas/<name>/{SOUL,MEMORY,USER}.md`
- Resolution centralized in `SystemPromptBuilder._resolve_persona`, so every caller (CLI, server) benefits from one code path
- Path-traversal guard on persona names (rejects `..`, `/`, `\`, leading `/`)
- **Default (empty `persona_name`) preserves today's global-persona behavior exactly** — no change for existing users

Fixes #380.

## Tests

- 8 new in `tests/prompt/test_persona_scope.py`: empty → passthrough, `none` → opt-out, named → personas dir, parametrized path-traversal rejection
- 21 existing prompt/persona tests pass; `ruff` clean

## Context

Developed as the live test case for the harnessability experiment in #475: a local 27B reasoner authored the spec independently from the issue + code — including the path-traversal guard — and the local squad implemented it. A follow-up adds the per-engine `[engine.personas]` extension (the multi-engine case) on request.

### DIFF
--- a/src/openjarvis/cli/ask.py
+++ b/src/openjarvis/cli/ask.py
@@ -151,9 +151,7 @@ def _run_research(
             trace.print(f"  [dim]↳ Found {n} {label}[/dim]")
         elif etype == "clarify_call":
             q = event.get("question", "") or ""
-            trace.print(
-                f"  [dim]↳ Clarifying:[/dim] [dim italic]{q}[/dim italic]"
-            )
+            trace.print(f"  [dim]↳ Clarifying:[/dim] [dim italic]{q}[/dim italic]")
         # final_answer and clarify_response are handled outside the loop.
 
     agent = ResearchAgent(
@@ -326,6 +324,7 @@ def _run_agent(
     temperature: float,
     max_tokens: int,
     capability_policy=None,
+    memory_files_config=None,
 ):
     """Instantiate and run an agent, returning the AgentResult."""
     # Import agents to trigger registration
@@ -374,7 +373,7 @@ def _run_agent(
 
         agent_kwargs["prompt_builder"] = SystemPromptBuilder(
             agent_template=config.agent.default_system_prompt or "",
-            memory_files_config=config.memory_files,
+            memory_files_config=memory_files_config or config.memory_files,
             system_prompt_config=config.system_prompt,
         )
 
@@ -592,6 +591,15 @@ def _print_profile(
         "(default: ~/.openjarvis/knowledge.db)."
     ),
 )
+@click.option(
+    "--persona",
+    "persona_name",
+    default=None,
+    help=(
+        "Named persona dir under ~/.openjarvis/personas/<name>/ "
+        "(overrides config). Pass 'none' to disable all persona files."
+    ),
+)
 @click.pass_context
 def ask(
     ctx: click.Context,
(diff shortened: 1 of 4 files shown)


buchkali:
Please stop all processes

robbym-dev (approved):
Approved: adds per-invocation --persona NAME/--persona none with a centralized _resolve_persona (path-traversal guarded). Empty persona_name preserves today's global-persona behavior exactly, so no change for existing users. CI green; 8 new tests. Closes #380.

PR #497, opened 2026-06-04
fix(chat): wire SystemPromptBuilder so persona files load in jarvis chat (fixes #458)

### LINKED ISSUE #458: fix(chat): wire SystemPromptBuilder so SOUL.md/MEMORY.md persona is loaded in `jarvis chat`
## Bug Description

`jarvis chat` does not load `SOUL.md` / `MEMORY.md` / `USER.md` persona files, even when `soul_path` is correctly set in `~/.openjarvis/config.toml`. `jarvis ask` works correctly because it explicitly wires `SystemPromptBuilder` — `chat_cmd.py` does not.

## Steps to Reproduce

1. Set `soul_path` in `~/.openjarvis/config.toml`
2. Run `jarvis ask "who am I?"` — persona is loaded correctly ✅
3. Run `jarvis chat` and ask the same — persona is **not** loaded ❌

## Root Cause

In `src/openjarvis/cli/ask.py`, `SystemPromptBuilder` is wired into the agent via `prompt_builder` kwarg:

```python
if "prompt_builder" in _inspect.signature(agent_cls.__init__).parameters:
    from openjarvis.prompt.builder import SystemPromptBuilder
    agent_kwargs["prompt_builder"] = SystemPromptBuilder(...)
```

`src/openjarvis/cli/chat_cmd.py` has no equivalent wiring, so agents instantiated by `jarvis chat` never receive the persona context.

Additionally, when no agent is used (direct engine mode), `chat_cmd.py` does not populate `system_prompt` from `soul_path`, so SOUL.md is never injected there either.

## Fix

Two changes needed in `chat_cmd.py`:

**1. Load SOUL.md as default system_prompt if none provided (after `config = load_config()`):**

```python
if system_prompt is None:
    import pathlib as _pl
    _soul = _pl.Path(getattr(config.memory, "soul_path", "~/.openjarvis/SOUL.md")).expanduser()
    if _soul.exists():
        _soul_txt = _soul.read_text().strip()
        if _soul_txt:
            system_prompt = _soul_txt
```

**2. Wire `prompt_builder` into agent kwargs (before `agent = agent_cls(engine, model, **kwargs)`):**

```python
import inspect as _inspect
if "prompt_builder" in _inspect.signature(agent_cls.__init__).parameters:
    from openjarvis.prompt.builder import SystemPromptBuilder
    kwargs["prompt_builder"] = SystemPromptBuilder(
        agent_template=getattr(config.agent, "default_system_prompt", "") or "",
        memory_files_config=getattr(config, "memory_files", None),
        system_prompt_config=getattr(config, "system_prompt", None),
    )
```

## Environment

- OpenJarvis v0.1.1 (also reproduced on latest main)
- Ubuntu 24.04 / WSL2
- `jarvis ask --tools ""` works; `jarvis chat` does not

SoulSniper-V2:
## Summary

Fixes Issue #458. Custom persona files (`SOUL.md`, `MEMORY.md`, `USER.md`) were previously ignored in the interactive `jarvis chat` REPL because `SystemPromptBuilder` was completely bypassed during initialization. This PR brings `chat_cmd.py` into parity with `ask.py` and the streaming server endpoints.

Additionally, this PR adds the `--persona` flag to `jarvis chat` to support the per-invocation scope overrides introduced in #380.

## What changed in `src/openjarvis/cli/chat_cmd.py`

1. **Added `--persona` flag**: Exposed the per-invocation override option to the CLI interface.
2. **Resolved `effective_mf`**: Replaced the global `config.memory_files` configuration gracefully if a persona override is provided.
3. **Agent Wiring**: Scanned `agent_cls.__init__` signatures and injected `SystemPromptBuilder` (populated with the effective memory configuration) exactly like `ask.py`.
4. **Direct Engine Wiring**: Invoked `builder.build()` for direct-to-engine chat mode so that bare engine queries correctly adopt the default or overridden persona.

## Test plan

- [x] Verified `tests/cli/test_chat_cmd.py` passes cleanly.
- [x] Ran full `ruff` formatter and linter checks cleanly.


gilbert-barajas (commented):
Reviewing as the author of the #380 design / #493 implementation this builds on — overall this is the right shape and brings `chat` to exact parity with `ask`. Verified against the PR head, not just the diff:

**Design lineage: correct.** `dataclasses.replace(config.memory_files, persona_name=...)` for the effective config, signature-scan `prompt_builder` injection mirroring `ask.py`, and the `--persona` / `'none'` sentinel surface matching what #493 actually shipped (I checked `ask.py` on main — there is no `--no-persona` flag, so this PR is consistent with the real surface, not the sketch in the #380 thread).

**Bonus catch, worth highlighting:** the `_load_file` empty-path guard fixes a latent bug in #493 itself — `--persona none` resolves to empty paths, and `Path("")` becomes `"."`, which exists, so the read raised `IsADirectoryError`. The original implementation (mine) missed it; the regression test locks it. Thank you.

**Verified no double-application:** in agent mode the REPL calls `agent.run(user_input)` without passing `history`, so the persona reaches the model exactly once (via the injected builder). No correctness issue.

**minor:** in agent mode the direct-mode fallback block (`if not system_prompt:` around L182) still executes — persona files are read and the prompt built a second time solely to seed `history`'s SYSTEM message, which the agent never consumes (it does make `/history` display a persona block the agent semantically has, so arguably intentional). If deliberate, a one-line comment would prevent a future "optimization" from changing UX; if not, gating it on `agent is None` saves the duplicate file reads.

**question (pre-existing pattern, not blocking):** precedence is asymmetric across modes — with both `--system` and an agent, the injected prompt_builder wins and `--system` is effectively ignored; in direct mode `--system` suppresses the persona build entirely. I believe this mirrors `ask.py`, so consistency wins either way; flagging only so the asymmetry is on record.

**nit:** the builder-level regression test is good; a CLI-level test asserting the first `history` message contains persona content in direct mode would lock the wiring this PR exists to fix.

No blocking concerns from my side — this completes the persona-scope table from #380 (`ask` ✅ / `agents` ✅ / `chat` ✅ with this).

PR #498, opened 2026-06-04
fix(leaderboard): correct telemetry pipeline and outlier handling

jonsaadfalcon:
## Summary

You flagged that the public leaderboard at \`/leaderboard\` showed a clear bimodal Wh/token distribution — most users at ~3-5 J/token, ~30% inflated by 1000-4000×. A 5-agent investigation workflow + 4-agent verification pass traced the inflation to a cluster of related bugs. This PR fixes the in-repo half. **Backfilling existing Supabase data is a separate follow-up that needs the service_role key** (proposed dry-run audit + additive migration outlined below).

## Bugs fixed

| # | Bug | Symptom | Fix |
|---|---|---|---|
| **1** | **Dual telemetry recording** in `routes.py:_handle_direct` — wraps an already-wrapped `InstrumentedEngine`, both layers fire `TELEMETRY_RECORD` | every chat-completion request counted 2× in leaderboard sums | Detect `InstrumentedEngine` and unwrap to `._inner` before passing to `instrumented_generate` |
| **2** | **Multi-turn fallback over-counts** in `savings.py:compute_savings` — falls back from `prompt_tokens_evaluated` to `prompt_tokens` (which is the per-turn sum of the full prompt) | FLOPs and energy estimates inflated by N× for N-turn conversations | Use `0` when the KV-cache-aware count is missing (conservative under-count) instead of the inflated sum |
| **3** | **No methodology versioning on telemetry records** | legacy + current records silently mixed in leaderboard sums even after upstream methodology changes | Add `token_counting_version` field + nullable column + idempotent migration; aggregator gains opt-in `current_methodology_only=True` filter that drops NULL rows for leaderboard queries (dashboards leave it off) |
| **4** | **Leaderboard JS displayed missing telemetry as legit zeros** | rows with tokens but no measured energy/FLOPs rendered as \`0.00 Wh\` — visually identical to a user who genuinely did almost nothing | New \`isMissingTelemetry()\` detector renders those cells as \`—\` with tooltip; new \`.lb-missing\` CSS distinguishes them |
| **5** | **Outlier bounds were 10,000× too generous** (\`MAX_ENERGY_WH_PER_TOKEN = 10\`) | every physically-impossible Group B row passed the filter | Tightened to \`0.5 Wh/token\` (still 500× over typical) + \`1e15\` FLOPs/token (still 10,000× over typical) — removes existing corrupt rows from the public view without touching Supabase |

## Tests

- \`tests/server/test_routes.py::test_instrumented_engine_unwrapped_to_avoid_dual_telemetry\` — asserts exactly ONE \`TELEMETRY_RECORD\` event per request when the engine is already an \`InstrumentedEngine\`. This would have caught Bug 1.
- \`tests/server/test_savings.py\` (NEW file, 3 tests) — pins Bug 2 fix (FLOPs not inflated via fallback) and the cost-side invariant (dollar savings still use full \`prompt_tokens\` because cloud providers bill per input token even with local KV cache hits).
- \`tests/telemetry/test_aggregator.py::TestMethodologyFilter\` (3 tests) — default includes legacy rows (dashboard parity); \`current_methodology_only=True\` excludes them; \`summary()\` honors the filter.

All 49 affected tests pass. Ruff clean.

## Unrelated test note

\`tests/telemetry/test_energy_wiring.py::TestTelemetryStatsEnergy::test_export_includes_energy_fields\` and \`::TestEndToEndPipeline::test_ask_to_export_with_energy\` are **flaky on this branch AND on \`main\`** (confirmed via stash + the banner-only run). Root cause: \`cli/_version_check.py:132-135\` writes the \"new version of OpenJarvis is available\" banner to stdout outside the throttle window, contaminating \`json.loads(result.output)\`. Reproducible with \`OPENJARVIS_NO_UPDATE_CHECK=1\` → all pass. Separate bug (the banner shouldn't write to the same stream as machine-readable output); not addressing here to keep this PR focused.

## Followup that needs Supabase access

After this lands, the natural cleanup is:

1. **Dry-run audit** (read-only): \`SELECT count(*) FROM savings_entries WHERE total_tokens > 1000 AND (energy_wh_saved / total_tokens) > 0.5;\` — verify the count is close to the expected ~30% of users before any write.
2. **Additive migration**: \`ALTER TABLE savings_entries ADD COLUMN methodology_version INTEGER DEFAULT 1;\` — existing rows = v1 (pre-fix); post-fix clients write v2.
3. **Filter the public fetch** in \`leaderboard.js\`: append \`&methodology_version=eq.2\` to the supabase select so the public view only renders post-fix entries. Historical data is preserved.
4. *(Optional, irreversible)* \`DELETE FROM savings_entries WHERE ...physically_impossible_bounds...\` if you want a clean slate.

Awaiting \`service_role\` key via secure delivery (\`umask 077 && echo '...' > ~/.supabase-srv-key\` or env var) to run any of these.

## What this PR explicitly does NOT do

- Backfill ~30% of Supabase rows that already shipped with inflated values from pre-fix clients.
- Distinguish which past submissions came from buggy vs correct clients retroactively (no \`app_version\` tag on existing submissions; new \`token_counting_version\` per-record column starts working only for post-fix submissions).

## Diff stat

\`\`\`
docs/javascripts/leaderboard.js                 | 44 +++++++++++--
docs/stylesheets/extra.css                      |  7 +++
src/openjarvis/core/types.py                    | 18 ++++++
src/openjarvis/server/routes.py                 | 22 ++++++-
src/openjarvis/server/savings.py                | 26 +++++---
src/openjarvis/telemetry/aggregator.py          | 38 ++++++++++--
src/openjarvis/telemetry/instrumented_engine.py |  9 ++-
src/openjarvis/telemetry/store.py               | 13 +++-
tests/server/test_routes.py                     | 51 ++++++++++++++++
tests/server/test_savings.py                    | 79 ++++++++++++++++++++++++
tests/telemetry/test_aggregator.py              | 58 ++++++++++++++++++
11 files changed, 343 insertions(+), 22 deletions(-)
\`\`\`

### DIFF
--- a/docs/javascripts/leaderboard.js
+++ b/docs/javascripts/leaderboard.js
@@ -12,8 +12,14 @@
   // Outlier detection — hide entries with values that are physically
   // implausible relative to their token count.  Thresholds are ~1000x
   // above legitimate per-token values to avoid false positives.
-  var MAX_ENERGY_WH_PER_TOKEN = 10;        // legit ≈ 0.001 Wh/tok
-  var MAX_FLOPS_PER_TOKEN = 1e17;           // legit ≈ 1e12 /tok
+  // Outlier bounds. Set well above realistic upper limits but tight
+  // enough to drop the pre-fix bimodal Group B (1-5 Wh/token, ~3e16
+  // FLOPs/token) — see the leaderboard PR for the full diagnosis.
+  // Realistic per-token rates on a consumer GPU + 10–30B local model:
+  // ~0.001 Wh/token, ~1e10–1e11 FLOPs/token. We allow 500× and 10,000×
+  // headroom respectively for inefficient hardware / larger models.
+  var MAX_ENERGY_WH_PER_TOKEN = 0.5;        // legit ≈ 0.001 Wh/tok
+  var MAX_FLOPS_PER_TOKEN = 1e15;           // legit ≈ 1e11 /tok
   var MAX_DOLLAR_PER_TOKEN = 25.0 / 1e6;   // hard ceiling: $25/1M output
 
   function isOutlier(row) {
@@ -27,6 +33,25 @@
       flops / tokens > MAX_FLOPS_PER_TOKEN ||
       dollars / tokens > MAX_DOLLAR_PER_TOKEN
     );
+  }
+
+  // Distinguish "user actually has zero work done" from "user's energy /
+  // FLOPs telemetry never landed". The latter happens when the server
+  // submits with valid dollar savings + token counts but the per-record
+  // energy stamp was missing (pre-fix builds, GPU energy meter
+  // unavailable, etc.). Without this check those rows show as "0.00 Wh"
+  // and skew the rankings + headline totals.
+  //
+  // Threshold: 1000 tokens is well above any single chat-turn — if a
+  // user has that many tokens recorded but no measured energy, the
+  // telemetry is incomplete, not legitimately zero.
+  var MIN_TOKENS_FOR_TELEMETRY = 1000;
+
+  function isMissingTelemetry(row) {
+    var tokens = Number(row.total_tokens) || 0;
+    var energy = Number(row.energy_wh_saved) || 0;
+    var flops = Number(row.flops_saved) || 0;
+    return tokens > MIN_TOKENS_FOR_TELEMETRY && energy === 0 && flops === 0;
   }
 
   function escapeHtml(s) {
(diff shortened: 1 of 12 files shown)


jonsaadfalcon:
**Heads-up: amended with a critical fix caught by adversarial review.**

The original implementation unwrapped `engine._inner` and passed the inner engine to `instrumented_generate`. The reviewer caught that this would have replaced "doubled records" with "records with no energy data AND no `token_counting_version`" — which the leaderboard's `current_methodology_only=True` filter would then drop entirely. Net effect: the leaderboard would have seen ZERO data from this path post-fix.

**Corrected fix**: when `engine` is already an `InstrumentedEngine`, **skip the wrapper entirely and call `engine.generate()` directly**. `InstrumentedEngine` publishes its own full record with energy + version intact. The lightweight `instrumented_generate` shim is only used when the engine isn't already instrumented.

**Strengthened test** (`test_instrumented_engine_unwrapped_to_avoid_dual_telemetry`) — now asserts:
1. Exactly ONE telemetry record per request (unchanged), AND
2. The surviving record carries `token_counting_version == TOKEN_COUNTING_VERSION` (so the leaderboard's methodology filter accepts it). The pre-amend implementation would have failed this second assertion.

Also tightened the misleading docstring in `savings.py` per the reviewer's finding 3.

All 27 affected tests pass.

PR #503, opened 2026-06-06
feat: add Whisper-based speech recognition with TTS test audio generation

Dantechz:
## Summary
- Added ackend/speech_recognition.py with OpenAI Whisper for speech-to-text transcription
- Integrated pyttsx3 for offline TTS test audio generation (run in subprocess to avoid COM/torch conflict)
- Used built-in winsound for audio playback on Windows, replacing pydub/simpleaudio which caused segfaults

## Test plan
- [ ] Run python backend/speech_recognition.py to verify test audio is generated
- [ ] Verify Whisper transcribes the generated audio correctly
- [ ] Confirm audio plays back without errors

?? Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/backend/speech_recognition.py
+++ b/backend/speech_recognition.py
@@ -1,0 +1,35 @@
+import subprocess
+import sys
+import winsound
+import whisper
+
+
+def generate_test_audio(text, output_path="test.wav"):
+    # Run pyttsx3 in a subprocess to avoid COM/torch conflict
+    script = (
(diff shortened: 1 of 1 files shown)


ElliotSlusky:
Closing after maintainer review. This is a small out-of-package Windows-only prototype with unconditional `winsound` usage and missing dependencies. Current main has integrated speech backends, so this is superseded.

PR #510, opened 2026-06-07
fix: isolate Google account sync and account-scoped retrieval

enkayxyz:
## What does this PR do?

Adds account-scoped Google connector support so multiple Google accounts can be connected side by side without sharing credentials, sync checkpoints, source IDs, or retrieval state.

Key changes:
- Store Google/Gmail provenance with account metadata and account-prefixed source IDs.
- Use account-scoped sync checkpoints such as `gmail:personal-main` instead of one shared `gmail` checkpoint.
- Support account-scoped retrieval filters such as `gmail:<account>`.
- Prevent non-empty searches from falling back to unrelated recent rows.
- Improve research behavior when account aliases or marker-like strings are accidentally treated as people filters.
- Add sync lookback handling so messages near checkpoint boundaries are not skipped.
- Document multi-account setup and account-scoped queries in the README.

## Why?

Users may have multiple Google accounts for personal, work, subscriptions, persona, or organization-specific data. Without strict account scoping, sync state and retrieval can produce misleading results or make it hard to prove which account a result came from.

This PR makes account provenance explicit and testable.

## How was this tested?

Automated local validation:

```bash
uv run --with pytest pytest tests/connectors/test_account_scoped_retrieval.py -q
# 4 passed

uv run --with ruff ruff check src/openjarvis/agents/research_loop.py src/openjarvis/connectors/sync_engine.py tests/connectors/test_account_scoped_retrieval.py
# All checks passed
```

Manual local validation with placeholder account aliases:
- Connected multiple Google/Gmail accounts using named account aliases.
- Ran clean Gmail syncs per account.
- Verified marker-like searches return results only for the requested `gmail:<account>` filter.
- Verified account/source isolation directly in SQLite using the invariant: `metadata.account` must match the `source_id` prefix.
- Verified mismatch count was `0` in the local test corpus.

## Known follow-up

Large multi-year Gmail corpora need better throttled sync orchestration, progress estimates, and transient network/backoff handling. This PR focuses on account isolation and retrieval correctness rather than full large-mailbox operations.

## Checklist

- [x] Tests pass for the touched account-scoping behavior
- [x] Linter passes for touched files
- [x] New/changed public behavior documented in README
- [x] Documentation updated where applicable

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,8 +73,15 @@ Example:
 ```bash
 jarvis init --preset morning-digest-mac
 jarvis connect gdrive          # one OAuth covers Gmail / Calendar / Tasks
+jarvis connect google --account work      # optional named Google profile
 jarvis digest --fresh          # generate and play your first briefing
 ```
+
+Named Google profiles let you connect accounts such as `personal`, `work`, or
+`subscriptions` side by side. Analysis can then stay scoped with prompts like
+`jarvis ask "Summarize only work Gmail about renewals"` or direct filters such
+as `source="gmail:work"`. See
+[Google Account Profiles](https://open-jarvis.github.io/OpenJarvis/user-guide/google-account-profiles/).
 
 Per-preset deep dives: [morning digest](https://open-jarvis.github.io/OpenJarvis/user-guide/morning-digest/) · [deep research](https://open-jarvis.github.io/OpenJarvis/user-guide/deep-research/) · [code assistant](https://open-jarvis.github.io/OpenJarvis/user-guide/code-assistant/) · [scheduled monitor](https://open-jarvis.github.io/OpenJarvis/user-guide/scheduled-monitor/) · [chat simple](https://open-jarvis.github.io/OpenJarvis/user-guide/chat-simple/) · or the full [quickstart guide](https://open-jarvis.github.io/OpenJarvis/getting-started/quickstart/).
 
--- a/README.md
+++ b/README.md
@@ -78,9 +78,36 @@ jarvis digest --fresh          # generate and play your first briefing
 ```
 
 Named Google profiles let you connect accounts such as `personal`, `work`, or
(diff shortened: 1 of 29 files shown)


enkayxyz:
Additional context from local validation:

This change was tested against multiple named Google/Gmail accounts with privacy-safe placeholder account aliases. The most important verification was at the storage/retrieval boundary: every Gmail chunk retained account provenance, and `metadata.account` matched the account-prefixed `source_id` for indexed Gmail rows.

The large-corpus testing also showed one useful follow-up: very large Gmail histories should use throttled/resumable sync with better progress estimates and network retry/backoff. I kept that as a follow-up because this PR is specifically about correctness of account isolation and account-scoped retrieval.

jonsaadfalcon:
Heads-up, @enkayxyz — we just merged #548, which fixes the Google Drive OAuth completion + Data-Sources registration bug (#512). #548 deliberately stayed out of `oauth.py` and kept the change OAuth-wiring-only to minimize collision with this PR.

Since this PR touches the same connector files, it'll need a rebase onto current `main`. The good news: your account-isolation work is orthogonal to the OAuth wiring and should rebase cleanly on top. One thing to watch during the rebase — please make sure #548's OAuth-completion fixes aren't reverted to the pre-fix versions (no background-browser thread in `handle_callback`; `Request` imported at module scope in `connectors_router.py`). Happy to help if anything's tricky — thanks for this work!

enkayxyz:
Thanks @jonsaadfalcon — rebased onto current `main` and preserved #548's OAuth fixes:
- `Request` stays at module scope in `connectors_router.py`
- No background-browser thread in the connector `connect()` paths
- Account/alias scoping re-applied on top

`tests/connectors/test_oauth_flow.py` + account-scoped tests all pass locally (52/52). Ready for another look when you have time!

CI workflows are awaiting maintainer approval on my fork PR — happy to address anything once checks run.

PR #523, opened 2026-06-10
fix(windows): installer encoding + Ollama readiness loop

### LINKED ISSUE #522: deploy/windows/install.ps1: two bugs breaking non-interactive and -OutFile installs on Windows
Two distinct bugs in `deploy/windows/install.ps1` surfaced during end-to-end testing on a clean **Windows 11 Pro 24H2 (build 26100)** Azure VM (`Standard_D4s_v5`). The canonical `irm … | iex` one-liner happens to dodge both, but the documented `-OutFile` fallback and any non-interactive run hit them.

## Bug 1 — UTF-8-without-BOM + em-dashes break the `-OutFile` install path on Windows PowerShell 5.1

`install.ps1` is encoded UTF-8 **without a BOM** and contains 21 em-dashes (`—`, U+2014) plus a box-drawing banner (lines ~490–492). Windows PowerShell 5.1 decodes BOM-less files with the legacy ANSI/OEM code page, not UTF-8, so the multi-byte sequences are mis-decoded and the parser desyncs — producing cascading `Missing closing ')'` / here-string parse errors (first reported around the Python-not-found here-string, lines ~159–212).

- ✅ `irm <url> | iex` works — `Invoke-RestMethod` decodes the HTTP body as UTF-8 before `iex` sees it.
- ❌ The README's documented fallback — `irm <url> -OutFile install.ps1; .\install.ps1` — **fails to parse**, because `-OutFile` writes raw bytes and `powershell -File` then reads them as ANSI.

**Repro:**
```powershell
irm https://open-jarvis.github.io/OpenJarvis/install.ps1 -OutFile install.ps1
.\install.ps1   # ParserError: Missing closing ')' ...
```

**Fix:** make the file pure ASCII (em-dashes → `-`, banner → ASCII art), or save it with a UTF-8 BOM.

## Bug 2 — Ollama daemon-readiness loop is dead code under `ErrorActionPreference='Stop'`

The script sets `$ErrorActionPreference = 'Stop'` near the top. The Ollama readiness loop (around line 343) probes with:

```powershell
& $ollamaExe list 2>&1 | Out-Null
if ($LASTEXITCODE -eq 0) { $ollamaReady = $true; break }
```

When the daemon isn't up yet, `ollama list` writes to stderr; `2>&1` merges that into the success stream where, under `Stop`, it becomes a terminating `NativeCommandError`. The script aborts (`exit 1`) on the **first** probe iteration — so the loop's own `Start-Process … serve` retry (at `$i -eq 5`) and the `Write-Warn2 "didn't become ready"` fallback are unreachable dead code.

Interactive installs usually dodge this because the Ollama tray app is already running (first `ollama list` returns 0, no stderr), but any install where the daemon isn't already up — including headless/non-interactive runs — aborts here instead of self-starting the daemon.

**Fix:** wrap the probe in `try/catch` (gating on `$LASTEXITCODE`), or drop the `2>&1` merge so a not-yet-ready daemon doesn't terminate the script.

---

Both found during end-to-end testing on a clean Windows 11 24H2 Azure VM. Fix incoming on `fix/windows-installer`.

robbym-dev:
Fixes #522.

Two bugs in `deploy/windows/install.ps1`, both found during end-to-end testing on a clean **Windows 11 Pro 24H2 (build 26100)** Azure VM. The canonical `irm … | iex` one-liner dodges both; the documented `-OutFile` fallback and any non-interactive run hit them.

## 1. Encoding - `-OutFile` path unparseable on Windows PowerShell 5.1
The script was UTF-8 **without a BOM** and contained 21 em-dashes (`—`) plus a box-drawing banner. PS 5.1 decodes BOM-less files with the legacy ANSI/OEM code page, mis-decoding those multi-byte sequences and desyncing the parser into cascading `Missing closing ')'` / here-string parse errors.

**Fix:** converted the file to pure ASCII - em-dashes -> `-`, banner box -> `+` / `-` / `|`. No BOM dependency; it now parses identically however it is read (`iex`, `-File`, or `-OutFile` then run).

## 2. Ollama readiness loop was dead code under `$ErrorActionPreference='Stop'`
The probe `& $ollamaExe list 2>&1 | Out-Null` merged the daemon-not-up stderr into the success stream, where `Stop` turned it into a terminating `NativeCommandError` — aborting the install on the **first** iteration, so the loop's own `Start-Process … serve` retry and `Write-Warn2` fallback never ran.

**Fix:** wrapped the probe in `try/catch` (gating on `$LASTEXITCODE`) so a not-yet-ready daemon falls through to the self-start path as designed.

## Testing
- During the original VM install both bugs were observed live: the raw `-OutFile` run produced the here-string parse errors, and the install aborted at the Ollama `list` probe. Re-saving with a BOM let it run clean through `uv sync --extra server` (full Windows dep tree + `openjarvis==0.1.1`); smoke tests passed (`jarvis --help`, `import openjarvis`, `import openjarvis.server.app`).
- Diff is `install.ps1`-only; pure-ASCII verified (`grep -P '[^\x00-\x7F]'` → empty).
- A formal re-verification on PS 5.1 (parse the `-OutFile`'d branch file; demonstrate the `try/catch` survives a failing probe under `Stop`) will be posted as a comment.

_Scope note:_ `deploy/windows/jarvis-service.ps1` carries the same em-dash encoding issue (it's also run via `-File`); left for a focused follow-up to keep this PR matched to #522.

### DIFF
--- a/deploy/windows/install.ps1
+++ b/deploy/windows/install.ps1
@@ -5,12 +5,12 @@
 .DESCRIPTION
     Phase-1 of the native-Windows-support RFC (#298). Mirrors the
     behavior of scripts/install/install.sh (the curl-pipe-bash installer
-    for Linux/WSL2/macOS) but for native Windows PowerShell — no WSL,
+    for Linux/WSL2/macOS) but for native Windows PowerShell - no WSL,
     no Docker, no MSYS2.
 
     Steps:
       1. Refuse non-Windows / Windows < 10.
-      2. Check Python 3.10 — 3.13 on PATH (3.14 has no numpy wheels yet,
+      2. Check Python 3.10 - 3.13 on PATH (3.14 has no numpy wheels yet,
          see #432).
       3. Check git on PATH.
       4. Install uv (https://astral.sh/uv) if absent.
@@ -65,7 +65,7 @@ if (-not $Service     -and $env:OPENJARVIS_SERVICE)      { $Service     = $true 
 if (-not $Force       -and $env:OPENJARVIS_FORCE)        { $Force       = $true }
 
 # ---------------------------------------------------------------------------
-# Output helpers — coloured but plain enough for Constrained Language Mode.
+# Output helpers - coloured but plain enough for Constrained Language Mode.
 # ---------------------------------------------------------------------------
 
 function Write-Info  ($msg) { Write-Host "[info]  $msg" -ForegroundColor Cyan }
@@ -77,13 +77,13 @@ function Write-Fail  ($msg) {
 }
 
 # ---------------------------------------------------------------------------
-# Shared helpers — winget bootstrap + PATH refresh
+# Shared helpers - winget bootstrap + PATH refresh
 # ---------------------------------------------------------------------------
 
 # Pull the latest Machine + User PATH from the registry into the current
 # PowerShell session. Tools installed by `winget install` (Python, git,
 # Ollama, etc.) update the User PATH, but the running process inherits
-# the parent shell's environment — so without this refresh the just-
+# the parent shell's environment - so without this refresh the just-
 # installed tool stays invisible to subsequent `Get-Command` calls.
 #
 # CRITICAL: registry PATH entries can be REG_EXPAND_SZ (with literal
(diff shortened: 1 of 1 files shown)


robbym-dev:
Verified on the Azure VM (PS 5.1.26100.8457, Win 11 24H2):

**Bug 1 (encoding):** Downloaded this branch's install.ps1 via 
-OutFile (no BOM), parsed with [Parser]::ParseFile - 0 errors, 
0 non-ASCII bytes. Previously blew up with cascading parse errors.

**Bug 2 (Ollama loop):** Under $ErrorActionPreference='Stop':
- Old bare `& list 2>&1` aborts immediately (confirmed)
- try/catch version: daemon down → doesn't abort, exit -1, 
  loop continues to the serve retry
- daemon up, exit 0, breaks as ready

Full install worked end-to-end: uv sync, jarvis --help, 
import openjarvis, Ollama inference with qwen2.5:0.5b all pass.

PR #536, opened 2026-06-11
fix(evals): harden terminalbench-native harness — tmux death + setup-hang misattribution

jonsaadfalcon:
## Root causes

**1. Terminus/tmux death leaks the container and dies opaquely**
`TerminalBenchTaskEnv.__enter__` (`src/openjarvis/evals/execution/terminalbench_env.py:51-63` on main) called `create_session` *after* entering the `spin_up_terminal` generator-CM with no exception safety. On a tmux failure the env's `__exit__` never ran: `docker compose down` was deferred to GC (never, if the env was retained), and the third-party error propagated with no task/image context — an opaque mid-run death with the GPU still burning.

**Fix:** exception-safe `__enter__` with an idempotent `_teardown()` (secondary cleanup errors logged, never masking the original; `KeyboardInterrupt`/`SystemExit` cleaned up but never wrapped), a **tmux + asciinema preflight** inside the container *before* the agent loop that raises a `TaskEnvironmentError` naming the task id, task image, container, exit code and remedy, and **fail-that-task-cleanly semantics**: the `AgenticRunner` records the failure as `QueryTrace.error_kind="harness_error"` (`agentic_runner.py`), the container is downed before the exception escapes, and the run continues with the next record.

**2. OpenHands SETUP hang recorded as a model result**
`TerminalBenchNativeBackend.run_harness` (`src/openjarvis/evals/backends/terminalbench_native.py:94-95` on main) called `harness.run()` with no bound — terminal-bench 0.2.18 runs installed-agent setup in-container with `max_timeout_sec=float("inf")` inside the per-trial agent budget, so a SETUP hang burned GPU indefinitely. Worse, `_run_terminalbench_native` (`src/openjarvis/evals/cli.py:691-711` on main) read the nonexistent `results.trial_results` attribute (the field is `results`) and hardcoded `errors=0`, so zero-model-request trials (`num_turns=1`, `is_resolved=false`, zero tokens) were silently folded into resolve-rate as model misses.

**Fix:** `global_agent_timeout_sec` / `global_timeout_multiplier` are threaded **eval config → backend kwargs → Harness kwargs** (new `RunConfig`/`ExecutionConfig`/`BenchmarkConfig` fields + TOML parsing; backend default `global_agent_timeout_sec=1800` bounds each trial's SETUP+RUN; `0` opts back into per-task budgets; an old terminal-bench without the kwargs fails loudly instead of hanging silently or TypeError-ing opaquely). The summary conversion now goes through `summarize_benchmark_results()`, which fixes the `trial_results` attribute bug and classifies harness/infra failures out of the accuracy denominator into `RunSummary.errors` plus a loud per-task console listing. The same zero-model-contact detection runs on the `--agentic` path (no LM events + zero tokens → `harness_error`), and `export.py`/`_print_agentic_summary` exclude harness errors from resolve-rate while reporting them (`totals.harness_errors`, `harness_error_details`).

## Verifier corrections applied

- **No `terminal-bench` extra in pyproject** — terminal-bench requires Python >=3.12 while this project supports >=3.10,<3.14; an unmarked extra would break `uv lock` for 3.10–3.11. The dep stays undeclared (existing pattern); `pyproject.toml`/`uv.lock` are untouched. *Follow-up option:* a marker-gated extra, e.g. `eval-terminalbench = ["terminal-bench>=0.2.18,<0.3; python_version >= '3.12'"]`.
- **Zero-model-contact detection does NOT use `failure_mode`** — in terminal-bench 0.2.18 clean trials leave `failure_mode` UNSET on success *and* on genuine unresolved misses, so a `failure_mode`-based condition would flag every real model miss as a harness error and inflate accuracy. Detection keys on **token usage** instead (`total_input_tokens + total_output_tokens` zero/missing while unresolved ⇒ the agent never completed a model request — reliable for terminus-2, the only agent this backend selects, which reports real LiteLLM usage). Only the definitionally-infra modes `agent_installation_failed` / `unknown_agent_error` are additionally flagged. A genuine miss (tokens>0, `failure_mode` unset, `is_resolved=false`) stays in the denominator — covered by a mandatory regression test.
- `QueryTrace.to_dict()/from_dict()` are manual field lists, so the new `error`/`error_kind` fields are wired through both (plus backward-compat load test) — otherwise they would silently never reach `traces.jsonl`.

## Upstream follow-up

A true **setup-only** timeout for installed agents (OpenHands etc.) requires an upstream terminal-bench change: `abstract_installed_agent.py` runs `install-agent.sh` with `max_timeout_sec=float("inf")`. Until then, `global_agent_timeout_sec` bounds SETUP+RUN together (documented on the backend ctor). No upstream internals are monkeypatched here. Also note: terminal-bench's `asyncio.wait_for` cannot cancel the executor thread on expiry, so a hung in-container process is only cut off at container teardown — the bound makes expiry deterministic and correctly classified rather than instant.

## Tests (all mock-based, no Docker/terminal-bench needed — they run in CI)

`tests/evals/test_terminalbench_env.py` no longer importorskips terminal_bench (it used to silently skip in CI); it now fakes the module tree in `sys.modules`. New: `tests/evals/test_terminalbench_native_backend.py`, `tests/evals/core/test_config_timeouts.py`; extended: `test_agentic_runner.py`, `test_export.py`.

- (a) tmux failure in `__enter__` → compose project downed before the exception escapes + actionable `TaskEnvironmentError` (+ teardown idempotence, compose-stderr surfacing, KeyboardInterrupt not masked)
- (b) preflight catches missing tmux (exit 127) naming the task image, before `create_session`
- (c) one task's harness failure → recorded as `harness_error`, run continues, healthy record still processed
- (d) `global_agent_timeout_sec`/`global_timeout_multiplier` travel TOML → `RunConfig` → backend → `harness_kwargs` (defaults, overrides, `0`-disables, old-terminal-bench loud failure)
- (e) zero-model-contact trial → `harness_error`, excluded from accuracy; **mandatory control:** genuine model miss (tokens>0, `failure_mode` unset, unresolved) is NOT flagged (accuracy 0.5, not 1.0) — on both the native-harness path and the AgenticRunner path; plus `trial_results`-attr regression and trace round-trip/backward-compat

### Results — patched branch
```
$ PYTHONPATH=$PWD/src .venv/bin/python -m pytest tests/evals/test_terminalbench_env.py \
    tests/evals/test_terminalbench_native_backend.py tests/evals/test_agentic_runner.py \
    tests/evals/test_export.py tests/evals/core/test_config_timeouts.py -p no:cacheprovider -q
77 passed, 1 warning in 4.01s

$ PYTHONPATH=$PWD/src .venv/bin/python -m pytest tests/evals src/openjarvis/evals/tests \
    --ignore=tests/evals/comparison -p no:cacheprovider -q -m "not live and not cloud"
745 passed, 1 skipped, 1 warning in 127.11s
```
(`tests/evals/comparison` is excluded: it import-errors on missing optional `polars` on unmodified main too — pre-existing, unrelated.)

### Fail-on-unfixed — same tests against unmodified main `src/`
```
$ PYTHONPATH=/lambda/.../OpenJarvis/src .venv/bin/python -m pytest <same five targets> -q
ERROR tests/evals/test_terminalbench_env.py            (TaskEnvironmentError does not exist on main)
ERROR tests/evals/test_terminalbench_native_backend.py
ERROR tests/evals/test_agentic_runner.py
3 errors during collection

$ PYTHONPATH=/lambda/.../OpenJarvis/src .venv/bin/python -m pytest tests/evals/test_export.py \
    tests/evals/core/test_config_timeouts.py -q     # files importing only pre-existing symbols
FAILED ...test_export.py::TestHarnessErrorExclusion::test_summary_excludes_harness_errors_from_accuracy
FAILED ...test_export.py::TestHarnessErrorExclusion::test_efficiency_excludes_harness_errors
FAILED ...test_export.py::TestHarnessErrorExclusion::test_no_harness_errors_key_absent
FAILED ...test_export.py::TestTraceErrorFieldRoundTrip::test_round_trip_preserves_error_fields
FAILED ...test_export.py::TestTraceErrorFieldRoundTrip::test_old_trace_dicts_still_load
FAILED ...core/test_config_timeouts.py::TestTimeoutConfigPlumbing::test_run_level_timeouts_parse_and_expand
FAILED ...core/test_config_timeouts.py::TestTimeoutConfigPlumbing::test_benchmark_override_wins
FAILED ...core/test_config_timeouts.py::TestTimeoutConfigPlumbing::test_defaults_are_none
8 failed, 31 passed
```

`ruff check` + `ruff format --check` clean on all touched files (whole-file formatting deliberately not applied to `cli.py`, which is format-unclean on main outside my hunks — keeps the diff minimal for the sibling `fix/eval-external-endpoint` PR; the `api_base` line is untouched).

## Notes for reviewers
- Scoring-semantics change: harness errors are now excluded from the resolve-rate denominator (previously counted as model misses) and surfaced in `RunSummary.errors` / `totals.harness_errors` — raw per-trial statuses remain in artifacts/traces for audit.
- `QueryTrace` remains a plain `@dataclass` (no `slots=True`) on purpose — flipping it is out of scope per the existing file's form.
- Minimal-edit constraint honored in `_run_terminalbench_native`: only the timeout-kwargs threading and the (load-bearing) `trial_results`→`results` summary conversion were touched; `api_base` line untouched.


brandonoelze49-gif (commented):
Probieren geht über studieren

PR #537, opened 2026-06-12
Fix engine resolution in ask and chat commands (fixes #524)

### LINKED ISSUE #524: Extend configured-engine resolution to `jarvis ask` / `jarvis chat` (not just serve)
Follow-up to #463 (which fixes engine resolution for the **server** path only).

PR #463 makes the server's lightweight system resolve the user's configured engine (`intelligence.preferred_engine` → `engine.default`) via `get_engine`, instead of hardcoding/falling through to a cloud client. However, the **CLI `ask` and `chat` paths still have the original problem** — this is the root cause of:

- #520 — `jarvis ask`/`chat` fail with "OpenAI client not available" even though a local Ollama is reachable
- #335 — same symptom on Windows with local Ollama

### What to do
Apply the same configured-engine resolution used in `_make_lightweight_system` (after #463) to `src/openjarvis/cli/ask.py` and `src/openjarvis/cli/chat_cmd.py` so a healthy local engine is selected before any cloud fallback, and a key-less `CloudEngine` is never reached on the happy path.

### Acceptance
- With `engine.default = "ollama"` (reachable) and no cloud keys, `jarvis ask "hi"` and `jarvis chat` use Ollama and do not raise "OpenAI client not available".
- Regression tests covering the CLI engine-selection path (mockable, no live engine).

Surfaced by a multi-agent triage of the open issues/PRs.

SoulSniper-V2:
## What does this PR do?

This PR fixes the engine resolution in the CLI `ask` and `chat` commands to use the standard fallback logic matching `_make_lightweight_system`. It correctly cascades from the explicit CLI `engine_key` to `config.intelligence.preferred_engine` and finally to `config.engine.default`.

Closes #524

## How was this tested?

- Added `tests/cli/test_ask_engine_resolution.py` to cover all three fallback scenarios.
- Ran the full test suite (`uv run pytest tests/ -v`).

## Checklist


jonsaadfalcon:
Thank you for this, @SoulSniper-V2 — and for the run of solid engine/chat fixes you've sent (#497, #514 too). 🙏

After digging in: the core problem this targets (a key-less cloud engine being picked as a fallback for `ask`/`chat`, surfacing as "OpenAI client not available") was already resolved on `main` by #533, which added `CloudEngine.can_serve()` + model-aware `get_engine()`. We verified the `ask`/`chat` paths now select a healthy local engine correctly, so most of this diff is a no-op against current `main`, and the engine-selection behavior your change guards is now covered by the regression tests added in #533/#545.

Closing as superseded by #533 — but genuinely grateful for the careful diagnosis. Please keep them coming!

SoulSniper-V2:
> Thank you for this, @SoulSniper-V2 — and for the run of solid engine/chat fixes you've sent (#497, #514 too). 🙏
> 
> After digging in: the core problem this targets (a key-less cloud engine being picked as a fallback for `ask`/`chat`, surfacing as "OpenAI client not available") was already resolved on `main` by #533, which added `CloudEngine.can_serve()` + model-aware `get_engine()`. We verified the `ask`/`chat` paths now select a healthy local engine correctly, so most of this diff is a no-op against current `main`, and the engine-selection behavior your change guards is now covered by the regression tests added in #533/#545.
> 
> Closing as superseded by #533 — but genuinely grateful for the careful diagnosis. Please keep them coming!

No worries at all. Glad the core issue got resolved in #533. I'll take a look at the issue tracker and see what else I can help with!

Issue #520, opened 2026-06-10
ask/chat fails with "OpenAI client not available" even though ollama engine is reachable


Azatik2008:
## Summary
`jarvis ask` and `jarvis chat` fail with an OpenAI client error, even though `jarvis doctor` confirms the ollama engine is reachable and models are ready. Talking to Ollama directly (`ollama run qwen3.5:4b`) works fine, so the local model itself is healthy — the failure is in OpenJarvis's engine routing.

## Error
Engine error: OpenAI client not available — set OPENAI_API_KEY and install openjarvis
Hint: Engine 'ollama' is not reachable.

## Environment
- Version: v0.1.1 (install method: editable-git)
- Platform: macOS (darwin), Apple M2, 8GB RAM
- Python 3.13.13

## Config
- engine.default = "ollama"
- intelligence.default_model = "qwen3.5:4b"
- provider = "ollama"

## What `jarvis doctor` shows
- Engine: ollama → Reachable
- Models: ollama → qwen3.5:4b, qwen3.5:2b
- Default model → "ollama/qwen3.5:4b not found on any engine" (warning)
- Also: REST API server "Not installed (openjarvis)"

## What I've tried
- Set engine to local, then ollama explicitly
- Tried model name with and without ollama/ prefix
- Ran jarvis quickstart (step 3 reports "Engine 'mlx' not reachable; falling back to cloud")
- jarvis self-update and manual git pull origin main both say "Already up to date" but version stays 0.1.1
- Confirmed ollama run qwen3.5:4b works directly

## Expected
jarvis ask should route to the reachable ollama engine instead of falling back to an OpenAI/cloud client.

### Steps to Reproduce

1. Install OpenJarvis on macOS (Apple Silicon)
2. Set config: engine.default = "ollama", default_model = "qwen3.5:4b", provider = "ollama"
3. Confirm `jarvis doctor` shows ollama engine Reachable and models ready
4. Run: jarvis ask "hello"

### Expected Behavior

jarvis ask should route the query to the local ollama engine (which doctor confirms is reachable) and return a response from qwen3.5:4b.

### Actual Behavior

The query fails with an OpenAI client error and a hint that the ollama engine is not reachable, even though `jarvis doctor` reports it as Reachable. Running `ollama run qwen3.5:4b` directly works fine, so the local model is healthy — the issue is in OpenJarvis engine routing falling back to a cloud/OpenAI client.

### Operating System

macOS

### Python Version

3.13

### Hardware

[...]

Issue #478, opened 2026-06-02
Jarvis version unchanged after successful update


selena731:
### Description

```
selen@DESKTOP-GG9Q00M:~/OpenJarvis$ uv run jarvis self-update
Current OpenJarvis version: v0.1.1
Install method: editable-git
Upgrade command: cd /home/selen/OpenJarvis && git pull && uv sync

Run the upgrade command now? [Y/n]:

→ cd /home/selen/OpenJarvis && git pull && uv sync

Already up to date.
Resolved 383 packages in 1ms
Uninstalled 4 packages in 32ms
 - fastapi==0.129.0
 - python-multipart==0.0.22
 - starlette==0.52.1
 - uvicorn==0.41.0

Upgrade complete. Re-run `jarvis --version` to confirm.
selen@DESKTOP-GG9Q00M:~/OpenJarvis$ uv run jarvis --version
jarvis, version 0.1.1
selen@DESKTOP-GG9Q00M:~/OpenJarvis$ git pull
Already up to date.
selen@DESKTOP-GG9Q00M:~/OpenJarvis$ uv sync
Resolved 383 packages in 2ms
Checked 61 packages in 14ms
selen@DESKTOP-GG9Q00M:~/OpenJarvis$ uv run jarvis doctor
A new version of OpenJarvis is available (v0.1.1 → v1.0.2)
Update: cd /home/selen/OpenJarvis && git pull && uv sync
Or run: jarvis self-update


OpenJarvis Doctor
....
```

### Steps to Reproduce

Run update commands (self-update or separately)

### Expected Behavior

New version warning does not show

### Actual Behavior

Version still remains the same

### Operating System

Windows

### Python Version

3.10

### Hardware

CPU only

### Engine

None

### Logs / Traceback

```shell

```

klxz79:
Yup. I can't use the desktop version because it keeps telling me there's an update available, I install the update but the version number doesn't change then the app tells me there's a new version again.

rdb16:
Root cause found (and workaround). The update notifier compares the installed version against the latest release tag (v1.0.2), but pyproject.toml on main still declares version = "0.1.1". So for editable-git installs, the suggested git pull && uv sync is a no-op by construction: main is up to date, yet it builds as 0.1.1.
Repro:
$ git pull          # → Already up to date (HEAD == origin/main)
$ grep -m1 version pyproject.toml   # → version = "0.1.1"
$ jarvis --version  # → jarvis, version 0.1.1
[...]

PR #538, opened 2026-06-12
Adopt hatch-vcs dynamic versioning (reworks autotag, pypi-publish, desktop)

### LINKED ISSUE #526: Adopt hatch-vcs dynamic versioning (reworks autotag + pypi-publish)
Follow-up to #525, which restored `pyproject` `version` to `1.0.2` (stopping the active tag/PyPI regression). That fixes reporting for **installed** users, but an **editable git checkout** still reports the static `1.0.2` rather than its true `git describe` version.

### Goal
Adopt **hatch-vcs** so the version is derived from git tags, and `jarvis --version` reflects the exact build (e.g. `1.0.3.dev750+g<sha>`) for source/editable installs.

### Why it's not a one-liner (the pipeline conflict)
- `pyproject.toml` would switch to `dynamic = ["version"]` + `[tool.hatch.version] source = "vcs"` — **removing** the static `version = "..."` line.
- `pypi-publish.yml` currently injects the version with `sed -i 's/^version = "..."/.../'` from the tag at build time — this breaks when there's no static line. It must instead let hatch-vcs derive the version from the checked-out tag.
- `autotag.yml` currently computes the next dev version by reading `pyproject`'s `version`; it must instead derive the base from the latest git tag (`git describe`).
- Need a clean **1.0.x anchor tag** reachable from `main` so hatch-vcs derives `1.0.3.dev*` (not the stray `0.1.2.dev*` tags).

### Risk
Touches the live PyPI + desktop release path; validate carefully (ideally a dry-run publish to TestPyPI) before merging.

Refs #478, #520, #525.

alarcritty:
## Problem

Closes #526.

#525 restored `version = "1.0.2"` in pyproject, which fixes version reporting for installed users. But an editable/source git checkout still reports the static `1.0.2` instead of its true `git describe` version, so `jarvis --version` on a dev checkout is misleading.

The version is also injected by hand in several places that all depend on that static line:
- `pypi-publish.yml` seds the static line from the tag at build time.
- `autotag.yml` computes the next dev version by grepping the static line.
- `desktop.yml` has a `workflow_dispatch` fallback that does the same grep.

Removing the static line (required to adopt hatch-vcs) breaks all three.

## Solution

Adopt hatch-vcs so the version derives from git tags, and rework the consumers that depended on the static line.

### pyproject.toml
- `dynamic = ["version"]` + `[tool.hatch.version] source = "vcs"`.
- Editable/source builds now report e.g. `1.0.3.dev110+g<sha>`.
- Exclude `.dev`/`.rc`/`desktop-*` tags from derivation. setuptools_scm cannot bump custom `.devN` tags, so the base is the latest plain release tag (`vX.Y.Z`) and the dev distance is the commit count since it.
- `fallback_version = "0.0.0+unknown"` so git-less builds (shallow CI clones, Docker `COPY src/`, source zips) resolve to a sentinel instead of hard-failing.

### pypi-publish.yml
- Drop the `sed` injection. Pin the exact build version from the pushed tag via `SETUPTOOLS_SCM_PRETEND_VERSION`, so the published version equals the tag.
- This pin is required, not cosmetic: a naive hatch-vcs build emits `1.0.3.devN+g<sha>`, and PyPI rejects local version segments on upload.
- Add a `dry_run` input that publishes to TestPyPI instead of PyPI, for validating the release path without a production upload.

### autotag.yml
- Derive the dev base from the latest release tag instead of grepping the (now removed) static pyproject version. The produced tag (`vX.Y.Z.devN`) is unchanged.

### desktop.yml
- The `workflow_dispatch` fallback grepped the same static line; derive its base from the latest release tag too, and give the build-and-release checkout full history and tags so the derivation works.

## Verification

- Editable install reports the derived version end to end: `importlib.metadata.version("openjarvis")` returns `1.0.3.dev110+g<sha>` (the path `jarvis --version` uses), not the static `1.0.2`.
- CI publish path: `SETUPTOOLS_SCM_PRETEND_VERSION=1.0.3.dev826 uv build` produces `openjarvis-1.0.3.dev826.tar.gz` (clean, publishable).
- autotag and the desktop fallback both compute `1.0.3.dev<count>`, matching the existing tag sequence.
- Targeted tests pass (docker structure, CLI version, SDK version); `ruff check` is clean.

## Notes and tradeoffs

- A clean `1.0.x` anchor tag already exists (`v1.0.2`), and dev tags are excluded from derivation, so the stray `v0.1.2.dev*` tags can never be picked. No new anchor tag is needed.
- Local-editable dev numbers are distance-based (`dev110`) while published/tag numbers are commit-count-based (`dev826`). They never collide because local builds are not published.
- TestPyPI dry run (maintainer step): `gh workflow run pypi-publish.yml --ref v1.0.3.dev826 -f tag=v1.0.3.dev826 -f dry_run=true`. With no `TEST_PYPI_API_TOKEN` it builds and skips the upload; add the secret for a real TestPyPI upload.

Refs #478, #520, #525.


jonsaadfalcon:
Merged — thank you, @alarcritty! 🙏 This was a careful piece of release engineering: the dev-tag base derivation, the no-`.git` Docker `fallback_version`, and the `SETUPTOOLS_SCM_PRETEND_VERSION` publish path were all handled well. We verified it end-to-end on the live pipeline — the first dynamic-version build autotagged `v1.0.3.dev834` and published to PyPI cleanly. Closes #526.

alarcritty:
> Merged — thank you, @alarcritty! 🙏 This was a careful piece of release engineering: the dev-tag base derivation, the no-`.git` Docker `fallback_version`, and the `SETUPTOOLS_SCM_PRETEND_VERSION` publish path were all handled well. We verified it end-to-end on the live pipeline — the first dynamic-version build autotagged `v1.0.3.dev834` and published to PyPI cleanly. Closes #526.
> 

Thanks for the thorough review and kind words! Really glad the end-to-end verification went smoothly and v1.0.3.dev834 landed cleanly on PyPI. Happy to help if anything else comes up down the line. 🚀

PR #576, opened 2026-06-21
fix(research): respect configured engine/model in Deep Research

### LINKED ISSUE #575: Deep Research ignores configured engine; hardcodes Ollama
### Description

Deep Research is hard coded to use Ollama.  If your main inference engine is your own trained model being served from LM Studio, vLLM or LlamaServer, Research will still try to be done using Ollama

### Steps to Reproduce

1. Run a Deep Research prompt without Ollama gemma4:31b running locally
2. Observe the 404 failure response

### Expected Behavior

I would expect the model that I'm already using for Chat to also be used for Deep Resarch as selected in the upper left

### Actual Behavior

Fails with a 404 stating the model cannot be found.

Image

### Operating System

macOS

### Python Version

3.12

### Hardware

Apple Silicon

### Engine

Other

### Logs / Traceback

```shell
Research failed: RuntimeError: Ollama returned 404: {"error":"model 'gemma4:31b' not found"}
```

chip902:
Fixes #575.

## Problem

Deep Research is hardwired to Ollama. `server/research_router.py` constructs `OllamaEngine()` directly and always runs the planner on `DEFAULT_PLANNER_MODEL = "gemma4:31b"`, ignoring the configured inference engine. On a non-Ollama setup (e.g. a model served from LM Studio / vLLM / llama-server) Deep Research fails with:

```
Research failed: RuntimeError: Ollama returned 404: {"error":"model 'gemma4:31b' not found"}
```

## Change

Resolve the Deep Research planner's engine/model from config instead of hardcoding it:

1. **Explicit override** — a new `[deep_research]` config section (`engine` / `model`).
2. **else the configured chat engine/model** — `engine.default` + `intelligence.default_model` (this is what the issue asks for: "use the model I'm already using for Chat").
3. **else the legacy** `ollama` / `gemma4:31b`.

The engine is built through the existing `get_engine()` registry; the `engine == "ollama"` path is unchanged. A new `DeepResearchConfig` dataclass is added (auto-included in the settable-config sections). The client-supplied `model` field keeps its existing "ignored" contract — the planner is server-resolved.

## Backwards compatibility

With no `[deep_research]` config, behavior is unchanged for Ollama users (it resolves to their configured engine/model, falling back to `gemma4:31b` only when nothing is set). No public APIs change.

## Tests

- `tests/core/test_deep_research_config.py` — the new config section + `validate_config_key` wiring.
- `tests/server/test_research_planner.py` — `_resolve_planner` for all three cases (legacy fallback, configured chat engine/model, explicit override).

Verified end-to-end: Deep Research now runs against an LM Studio engine over the network instead of 404-ing on Ollama.

### DIFF
--- a/tests/server/test_research_planner.py
+++ b/tests/server/test_research_planner.py
@@ -12,10 +12,10 @@ def test_legacy_fallback_when_nothing_configured():
 def test_uses_configured_chat_engine_and_model():
     cfg = JarvisConfig()
     cfg.engine.default = "lmstudio"
-    cfg.intelligence.default_model = "qwen3.5-9b-uncensored-hauhaucs-aggressive"
+    cfg.intelligence.default_model = "fine-tuned-model"
     assert _resolve_planner(cfg) == (
         "lmstudio",
-        "qwen3.5-9b-uncensored-hauhaucs-aggressive",
+        "fine-tuned-model",
     )
 
 
--- a/tests/server/test_research_planner.py
+++ b/tests/server/test_research_planner.py
@@ -1,0 +1,28 @@
+from openjarvis.core.config import JarvisConfig
+from openjarvis.server.research_router import _resolve_planner
+
+
+def test_legacy_fallback_when_nothing_configured():
+    cfg = JarvisConfig()
+    cfg.engine.default = "ollama"
+    cfg.intelligence.default_model = ""
+    assert _resolve_planner(cfg) == ("ollama", "gemma4:31b")
+
+
+def test_uses_configured_chat_engine_and_model():
+    cfg = JarvisConfig()
+    cfg.engine.default = "lmstudio"
+    cfg.intelligence.default_model = "qwen3.5-9b-uncensored-hauhaucs-aggressive"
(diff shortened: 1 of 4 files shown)


tests/server/test_research_planner.py
```diff
@@ -0,0 +1,28 @@
...
+    cfg = JarvisConfig()
+    cfg.engine.default = "ollama"
+    cfg.intelligence.default_model = ""
+    assert _resolve_planner(cfg) == ("ollama", "gemma4:31b")
+
+
+def test_uses_configured_chat_engine_and_model():
+    cfg = JarvisConfig()
+    cfg.engine.default = "lmstudio"
+    cfg.intelligence.default_model = "qwen3.5-9b-uncensored-hauhaucs-aggressive"
```
ElliotSlusky: Why this model?

chip902: @ElliotSlusky It doesn't have to be that one.  just happens to be what I use in LMStudio.  I updated the test

PR #586, opened 2026-06-23
 feat(connectors): generic IMAP email connector for any provider (#518)

### LINKED ISSUE #518: Custom Email connector
### Problem Statement

Not having Outlook nor Gmail account

### Proposed Solution

Rely on SMTP/IMAP connection to connect to any inbox 

### Primitive Area

Channels

### Alternatives Considered

https://github.com/pewdiepie-archdaemon/odysseus

alarcritty:
Problem

  Closes #518.

  OpenJarvis only had provider-specific email connectors (Gmail OAuth, Gmail IMAP, Outlook), each hardcoded to one
  provider's IMAP host. A user without a Gmail or Outlook account had no way to index their mailbox. The issue
  asks to rely on SMTP/IMAP to connect to any inbox.

  Testing this end to end against a real inbox surfaced two adjacent problems that broke the flow: a
  header-encoding crash during indexing, and Deep Research being pinned to a model that was not pulled.

  Solution

  Generic IMAP connector (the core of #518)

  - New connectors/imap.py: one connector that works with any provider instead of a subclass per provider. The
  IMAP host is resolved from the email domain (a small table of common providers, falling back to imap.<domain>),
  so a user supplies only an email address and an app password.
  - Registered in connectors/__init__.py so it appears in the Data Sources API and UI.
  - New "Email (IMAP)" card in the frontend catalog (connectors.ts) with email and app-password inputs, alongside
  Gmail and Outlook.

  Reliability fixes in the shared IMAP sync

  - Decode every header (From, To, Subject, In-Reply-To, Message-ID) to a plain string. Real mail with raw 8-bit
  header bytes made imaplib return an email.header.Header object, which crashed json.dumps in the knowledge store
  ("Object of type Header is not JSON serializable") and aborted the sync. The decoder also tolerates bogus
  charsets such as unknown-8bit. This also benefits the Gmail and Outlook IMAP connectors.
  - Reconnect and retry on a dropped TLS connection mid-backfill (for example [SSL: BAD_LENGTH]) instead of
  aborting the whole run, so a large inbox finishes without manual retries.

  Deep Research model resolution

  - server/research_router.py no longer hardcodes the planner model (which returned a 404 when that model was not
  pulled). It resolves from the server's running model, then the configured default, then the built-in fallback,
  so research runs on a model that is actually available.

  Chat tool wiring

  - cli/serve.py adds knowledge_search to the default chat agent tools and injects a KnowledgeStore, so
  non-streaming agent requests can search ingested data. Note: the streaming web chat path bypasses the agent, so
  this does not yet surface in the streaming UI.

  Verification

  - Tested against a real Gmail inbox: connected, backfilled, and indexed hundreds of emails into the knowledge
  store; Deep Research returns answers over the indexed mail.
  - New tests in tests/connectors/test_imap.py cover registration, domain host resolution, credential round-trip,
  document output, raw 8-bit headers, and reconnect-on-drop.
  - ruff check and ruff format clean; connector tests pass.

  Notes and limitations

  - Connects over implicit TLS on port 993 only. Providers that require STARTTLS on port 143 or a non-standard
  port, or whose IMAP host does not follow imap.<domain>, are not supported yet (there is no manual host/port
  field). Outlook or Microsoft 365 accounts with app-password IMAP disabled will not connect.
  - This branch bundles three concerns (IMAP connector for #518, a chat knowledge_search wiring change, and a Deep
  Research model fix), kept as separate commits. They can be split into separate PRs if preferred.


alarcritty:
@jonsaadfalcon please review.

ElliotSlusky:
Maintainer repair pushed in 8b22860f (append-only merge of current main; no force-push).

The generic connector now resolves the host from direct credentials as well as stored credentials, so direct Gmail and other provider setup cannot connect with an empty host. Generic results no longer carry Gmail URLs or advertise Gmail-only live tools. IMAP sessions now close on normal completion, generator cancellation, failed login/select, search failure, and reconnect paths; response statuses are also checked.

Validation:
- PYTHONPATH=src python3 -m pytest tests/connectors/test_imap.py -q: 12 passed
- ruff on the changed connector/tests: passed

ElliotSlusky (dismissed):

alarcritty:
hey @ElliotSlusky, do let me know if you need anything from my side.

PR #570, opened 2026-06-19
Fix: use httpx.AsyncClient in async FastAPI handlers to prevent event loop blocking (fixes #559)

SoulSniper-V2:
## What does this PR do?
This PR resolves issue #559 by converting several blocking, synchronous `httpx.Client()` calls inside `async def` FastAPI routes into their `httpx.AsyncClient()` equivalents. It also offloads the CPU-bound `backend.transcribe()` call in `api_routes.py` to a separate thread via `asyncio.to_thread()`. This prevents the Uvicorn event loop from freezing and causing poor latency for other parallel requests.

## Changes
- Replaced `httpx.Client` with `async with httpx.AsyncClient` in `server/routes.py` (`/v1/models/pull`, `/v1/models/{name}`).
- Replaced `httpx.get/post` with `async with httpx.AsyncClient` in `server/agent_manager_routes.py` (`/verify`, `/register-webhook`, `/test`).
- Offloaded Whisper transcription in `server/api_routes.py` to `asyncio.to_thread`.
- Added a regression test `tests/server/test_async_handlers_nonblocking.py` to verify event loop non-blocking behavior.

## Testing
- [x] Added unit/integration tests
- [x] Ran locally
- [x] Verified endpoints

## Fixes Issue
Fixes #559

PR #597, opened 2026-06-26
Fix: enable WAL and batch writes in TelemetryStore to avoid SQLITE_BUSY under concurrency (fixes #560)

### LINKED ISSUE #560: Fix: enable WAL and batch writes in TelemetryStore to avoid SQLITE_BUSY under concurrency
Enable WAL mode and batch commits in TelemetryStore to prevent SQLITE_BUSY and reduce I/O. 

Changes:
- Set PRAGMA journal_mode=WAL and PRAGMA synchronous=NORMAL when opening TelemetryStore connection
- Replace per-insert commit() with buffered/batched commits (e.g., commit every N inserts or time-based flush)
- Add unit tests to simulate concurrent writes and verify no SQLITE_BUSY raised
- Add migration notes if needed

SoulSniper-V2:
## What does this PR do?
Resolves Issue #560 and fully closes out parent Issue #219 alongside PR #570.

The `TelemetryStore` was previously missing `PRAGMA journal_mode=WAL` which made it susceptible to `SQLITE_BUSY` errors under concurrent inference load. Furthermore, every telemetry insert was immediately calling `commit()`, forcing high fsync I/O overhead.

## Changes
- **WAL Mode Enabled**: Set `PRAGMA journal_mode=WAL` and `PRAGMA synchronous=NORMAL` when initializing `TelemetryStore` so it correctly mimics `TraceStore`.
- **In-Memory Batching**: Added a `threading.Lock()` to `TelemetryStore` to buffer incoming records in `self._telemetry_batch` and `self._mining_batch`.
- **Reduced Disk I/O**: Modifed `record()` and `record_mining_stats()` to hold onto records until the `batch_size` (default: 50) is reached.
- **Flushing Behavior**: Added a thread-safe `flush()` method that uses SQLite's fast `executemany()` operation. This method is called automatically when hitting the batch limit, when querying (`list_recent`), and when closing the DB connection.
- **Unit Testing**: Added `test_batching_delays_commit` to `test_store.py` to ensure batching logic correctly defers disk commits.

## Testing
- [x] Added unit/integration tests
- [x] Ran locally
- [x] Verified endpoints

## Fixes Issue
Fixes #560


ElliotSlusky:
Thank you @SoulSniper-V2! Can you fix the tests and merge conflicts?

SoulSniper-V2:
Hey @ElliotSlusky! Fixed — merge conflicts are resolved and all 13 tests pass (including the new WAL and concurrency tests from main). I also integrated main's `busy_timeout=5000` alongside the batching logic. Ready for another look!

ElliotSlusky:
Pushed one more commit on top (hope that's okay — maintainer edits were enabled) addressing the last review notes:

- **Background flusher thread**: the stale-batch check only ran inside `record()` calls, so a partial batch written right before traffic stopped stayed invisible to readers on other connections (`TelemetryAggregator`, the leaderboard pipeline) until the *next* write — potentially forever on an idle server. A daemon thread now guarantees pending rows land within `flush_interval_seconds`; `flush_interval_seconds=0` disables time-based flushing entirely, which the batching test now uses for determinism.
- **Idempotent `close()`**: a second `close()` previously raised `ProgrammingError` (flush commits on a closed connection); it's now a guarded no-op, and the flusher thread shuts down cleanly with it.
- **Visibility contract documented** on the class docstring (when readers on separate connections see rows, and when to call `flush()`).
- Small test fix: `with sqlite3.connect(...)` is a transaction scope, not a close — swapped to `contextlib.closing` and added pins for the background flush and double-close.

The WAL + batching work itself was in great shape after your rebase — these were just the hardening leftovers. All 282 telemetry tests pass locally. Thanks for the quick turnaround on the earlier feedback @SoulSniper-V2!

SoulSniper-V2:
Thanks @ElliotSlusky! The background flusher is a great addition, the stale-batch edge case on idle servers was a blind spot on my end. Appreciate the hardening, everything looks good on my side and no conflicts with main.

ElliotSlusky (approved):
Re-verified against current main: clean preview merge, 282 telemetry tests pass, both Ruff gates clean. The flusher/close() shutdown ordering is correct and all closing consumers are wired.

PR #599, opened 2026-06-26
memory: provenance tagging + injection-scanner gate (prompt-injection defence)

TinkerOfThings:
## Summary

Hardens the automatic long-term-memory path against prompt injection. The fact extractor distils facts from raw chat exchanges that can contain hostile input (scraped pages, tool output, pasted content); those facts are stored and surfaced unfiltered. This adds provenance tagging, an injection gate around extraction, quarantine of facts that look hostile, and a review path to release one — plus it revives the injection scanner in Rust-less environments.

## Changes

- **Provenance** — `memory.store.Fact` gains a `trust` tier, round-tripped through JSONL. Three tiers: `auto` (extracted and scanner-clean), `trusted` (explicitly vouched for), `untrusted` (quarantined). Legacy rows carry `trust=""` and stay recallable, so upgrading never silently erases existing memory. Unknown future tiers fail closed rather than becoming prompt input. `FactStore.add`/`add_many` accept `trust`, and `FactStore.set_trust` promotes a fact.
- **Injection gate before extraction** — `MemoryService` scans each exchange with `InjectionScanner` **before** extraction, so an overt attempt never reaches the extraction model or the store. Only a **HIGH/CRITICAL** finding suppresses the exchange: dropping one is unrecoverable, and lower-severity patterns (a pasted shell one-liner, a role-delimiter fence quoted in a code discussion) are ordinary developer traffic. Scanning **fails open** — a scanner outage must not switch memory off.
- **Per-fact quarantine** — each extracted fact is then scanned on its own. A fact whose own text trips the scanner is stored `untrusted` and filtered out of every model-facing path (`load_configured_facts`, `inject_context`, `build_context_message`); clean facts from the same exchange stay recallable. Quarantining per fact rather than per exchange is what keeps automatic memory from becoming write-only.
- **Review path** — `jarvis memory list` shows a Trust column and flags quarantined rows; `jarvis memory trust <#>` promotes one after you've read it.
- **Scanner Python fallback** — `InjectionScanner` hard-required the `openjarvis_rust` extension and crashed on construction when it wasn't built. Added a pure-Python fallback over its own `_INJECTION_PATTERNS`, mirroring the `RUST_AVAILABLE` pattern used elsewhere (e.g. `security.ssrf`), and matched to the Rust backend's behaviour including the 100-character `matched_text` cap. This fixes 10 previously-failing scanner tests in Rust-less environments.

## Testing

- Test-first throughout, including an **end-to-end** test (service → store → `inject_context`) asserting a captured fact actually comes back out. Per-layer tests can each pass while the pipeline as a whole recalls nothing, so that one is the load-bearing check.
- Coverage for: provenance round-trip, legacy rows staying recallable, `auto` vs `untrusted` recall, per-fact quarantine with clean siblings, HIGH blocks / MEDIUM does not, fail-open, `set_trust` persistence, and the CLI quarantine marker and `memory trust` command.
- No regressions: full suite has the same 207 failures before and after (all pre-existing `openjarvis_rust`-extension and missing-optional-dependency failures); passing count 7093 → 7098. Affected scanner suites went 114 → 104 failing (10 fixed). `ruff check` clean.


ElliotSlusky:
Thanks so much for this PR — it's a genuinely thoughtful piece of work. The test-first approach shows, the provenance round-trip is clean, the fail-open reasoning is clearly documented in the code, and reviving the scanner in Rust-less environments (fixing 10 failing tests along the way) is a lovely bonus. I also went out of my way to verify a couple of things that came up while reading, and they both check out in your favor: the Python fallback patterns are byte-for-byte identical to the Rust set (so no detection weakening), and I timed the newly-activated Python regexes against crafted adversarial input — they scale linearly (~3ms on 30KB), so no ReDoS concerns. Nice.

A few things surfaced while tracing the code that I'd love your thoughts on, roughly in order of importance:

### 1. The quarantine doesn't currently reach the model-facing path

This is the big one, and it's more of a wiring/expectations issue than a bug in your code. Grepping the tree, `Fact.trust` is only ever *read* in the CLI table renderer (`cli/memory_cmd.py:200,204`). The memory that actually flows into LLM prompts is the separate `memory_backend` (`MemoryRegistry.create(...)` in `cli/serve.py:488`, consumed via `context_from_memory` in `system/orchestrator.py:43`) — and FactStore facts have no path into it. So the comment at `service.py:184` ("tag untrusted so recall/surfacing quarantines these facts") describes protection the runtime doesn't have yet: the tag warns a human running `memory list`, but the store that reaches the model is untouched by this PR.

Totally fine if the answer is "that's phase 2" — the provenance groundwork here is exactly what that would need. I'd just suggest either wiring the trust check into the recall path that feeds prompts, or softening the comment/PR description so future readers don't assume runtime containment that isn't there yet.

### 2. The gate may drop more benign exchanges than intended

`_exchange_is_malicious` (`service.py:167`) scans user *and* assistant text together and treats any finding at any threat level as malicious (`not result.is_clean`, where `is_clean = not findings`). So an entirely benign exchange — user asks "what is prompt injection?", assistant's answer quotes "ignore all previous instructions" — matches the `prompt_override` pattern and the whole exchange is skipped, with only a debug-level log. Since these facts would land as `trust="untrusted"` anyway, maybe gate only on HIGH/CRITICAL findings, or log at info/warning so the drops are visible?

### 3. Small regression in the tightened line fallback

The new marker regex requires whitespace after the bullet (`\s+`), where the old code stripped the marker with `\s*`. I ran the regex against a few cases:

```
'- User likes coffee'         -> 'User likes coffee'      ✓
'-User likes coffee'          -> DROPPED (old code kept it)
'1.Lives in Berlin'           -> DROPPED (old code kept it)
'1995. They live in Berlin.'  -> 'They live in Berlin.'   (prose still minted)
'User likes coffee'           -> DROPPED (intended, but silent)
```

Two suggestions: switching `\s+` back to `\s*` preserves your prose-rejection goal while keeping tightly-formatted bullets, and a debug log when the fallback rejects every line of a non-empty response would make "model drifted off the JSON spec" diagnosable rather than silently turning memory capture off. (The `1995.` case is inherent to accepting numbered lists at all, so probably just worth knowing about.)

### Smaller notes (take or leave)

- `store.py:147` — dedupe compares text only, so trust is frozen at first write: a legacy `trust=""` fact can never be re-quarantined if the same text is later re-derived from a hostile exchange. Bounded impact today (nothing in production writes trusted facts), but if quarantine becomes load-bearing, "colliding facts keep the least-trusted label" might be the safer rule.
- `cli/memory_cmd.py:200` — the display fails open: anything that isn't exactly `"untrusted"` renders green "trusted". A shared constant plus an inverted check (warn on anything not known-trusted) would fail safe if more tiers ever appear.
- `cli/memory_cmd.py:204` — tiny one: the untrusted predicate is computed in the row loop and again in a second full `any(...)` pass; a flag set inside the existing loop would cover both.

None of this diminishes the direction — provenance + gate + quarantine is the right shape, and this lands the foundations well. Happy to help with any of the follow-ups if useful. Thanks again!

TinkerOfThings:
Thanks Elliot — this is a fair review and the headline point lands. Going through it:

**On the trust tag having no runtime effect (your main point): you're right, and it's worse than a naming nit — the PR is incomplete against its own stated goal.**

The `trust="untrusted"` tag is currently inert at inference time because the store it's written to (`LocalFactStore`) is disjoint from the model-facing retrieval path. Auto-extracted facts go into `MemoryService`/`LocalFactStore`, which is read *only* by the `memory list` CLI. The recall path that actually reaches the model — `inject_context` over a `MemoryBackend` (BM25/SQLite) — never sees those facts and has no concept of trust. So the comment at `service.py` ("tag untrusted so recall/surfacing quarantines these facts") and the "prompt-injection defence" framing overclaimed a quarantine that isn't wired up. That's on me.

The intended design is exactly what the tag implies: auto-extracted facts *should* feed the model, with untrusted-origin ones filtered or down-weighted at recall. The machinery that's landed (injection-scanner gate, marker-only extraction, provenance tag) is the front half; the back half — the FactStore→recall bridge plus a trust filter — is missing. Concretely:

- `MemoryService` mirrors extracted facts into the `MemoryBackend` via `store(text, source="auto", metadata={"trust": "untrusted"})` (the backend already accepts arbitrary `metadata`, and it survives through to `RetrievalResult.metadata`);
- `inject_context`/`format_context` reads `metadata["trust"]` and drops or annotates untrusted results before they hit the model.

That closes the loop and makes the tag do real work. I'm tracking it as a follow-up rather than piling it into this PR — keeping this one scoped to the extraction/provenance layer and landing the bridge+filter separately with its own tests (recall must *not* surface an untrusted fact; a trusted one still does).

**On the skip-path being invisible (the debug-only log):** agreed — a silently-dropped exchange looked identical to "nothing to extract." Fixed in c0aa01b7: the skip log is now `info` and states the trigger. Separately, the gate is currently binary (any finding at any severity → skip); I'd like to gate on a severity threshold so a low-confidence hit doesn't nuke the whole exchange, but that's a tuning follow-up, not a blocker.

**On the line-fallback rejecting unmarked prose:** this one's intended, not a regression. Commit `62fd282e` tightened the fallback to require a list marker precisely so free-form model prose can't be laundered into "facts" without an explicit marker — markerless lines like `-User likes coffee` are dropped on purpose. The intent is already documented in-line at `extractor.py:130-133`.

**The smaller notes** (double-computed `untrusted` check in the CLI, overclaiming comment/docstring) are good catches — folded in with the same commit c0aa01b7: the CLI now computes the untrusted flag once, and the `service.py` comments no longer promise runtime quarantine.

Net: this PR stays scoped to the provenance/extraction layer with the overclaiming comments corrected so it doesn't promise runtime quarantine it doesn't yet deliver; I'll reframe the PR description in the same spirit, and open a follow-up for the bridge+filter that actually enforces trust at recall. Thanks for catching that the tag wasn't load-bearing yet.

TinkerOfThings:
Opened the follow-up as #632 (branch `memory-trust-recall-filter`), per the plan above.

It closes the loop this PR left open — `trust="untrusted"` is now load-bearing at inference time:

- `MemoryService` mirrors each newly-stored auto fact into the model-facing `MemoryBackend` as `metadata={"trust": "untrusted"}`.
- `apply_trust_policy` (wired into `inject_context`) drops untrusted results by default, or annotates them behind an unverified-origin caveat via `memory.context_untrusted_policy`.

Tests prove the property end-to-end, including a live mirror → recall run over the real sqlite backend: an untrusted fact does not surface at recall by default, a trusted one still does, and the annotate path surfaces it only behind the warning.

Keeping this PR scoped to the provenance/extraction layer.

ElliotSlusky:
Maintainer repair and current-main integration pushed in `6016f77`:

- untrusted and unknown provenance tiers are now excluded before recall budgeting and again during context-message construction
- `load_configured_facts()` filters quarantined facts for ask/SDK/system paths, while `FactStore.list()` retains them for audit
- legacy `source=auto` entries that predate trust tagging fail closed; legacy/manual entries remain compatible
- CLI and server/chat integration tests prove hostile auto-memory never reaches the engine prompt
- direct context-builder callers receive the same defensive filtering
- the Python injection-scanner fallback now has an explicit no-Rust regression test

Validation on current `main`: **139 focused tests passed** across memory, CLI chat, server recall, scanner, and Rust bridge suites; Ruff check/format passed on all changed files.

ElliotSlusky (dismissed):
Maintainer repair reviewed on current main: 139 focused tests pass. Provenance quarantine is now enforced across model-facing recall paths with end-to-end CLI/server assertions.

Issue #472, opened 2026-06-01
Optional on-device security scanning for imported skills before they run


theyavuzarslan:
### Problem Statement

Disclosing affiliation upfront: I work on AgentGuard at GoPlus Security (https://github.com/GoPlusSecurity/agentguard, MIT). Feature request, opening for discussion before any PR as your form intro suggests.

OpenJarvis installs skills from OpenClaw (~13,700 community skills), Hermes (~150), or any GitHub repo, following agentskills.io. Those skills then run locally with shell access, file I/O, code execution (native_openhands CodeAct), and OAuth'd access to Gmail / Calendar / Drive via the digest presets.
That's a large untrusted-code surface with a high blast radius on a personal device. A single malicious or compromised community skill could read secrets, exfiltrate personal data, or run arbitrary commands. Right now there's nothing between jarvis skill install and execution that inspects what a skill actually does. agentskills.io has no security gate, and 13,700 community skills can't be hand-vetted.

### Proposed Solution

Two optional checks, both fully on-device (no cloud call, no model call, so they don't break the local-first thesis):

Skill scan at install / sync time. Static analysis of each imported skill, SKILL.md included, for backdoors, obfuscation, secret reads, network and webhook exfiltration, and prompt injection, before it enters the catalog. jarvis skill install gets a scan gate.
Runtime tool-call guard (follow-up). A pre-execution check on shell / file / code_execution calls in the agent loop (orchestrator, native_openhands), returning allow / warn / block plus reason.

AgentGuard already does both. It ships an OpenClaw adapter and follows agentskills.io, the exact sources you import from, with 24 detection rules. Runs locally, sub-millisecond, no model call, so the overhead is negligible against your energy / FLOPs / latency budget. We already ship this as adapters for Claude Code (PreToolUse / PostToolUse) and OpenClaw (before_tool_call / after_tool_call) on a shared decision engine; an OpenJarvis adapter is the same pattern, third instance.
Why this fits OpenJarvis specifically:

Local-first: a cloud guardrail would contradict the entire premise. AgentGuard is on-device, no network round-trip, no model in the loop.
agentskills.io alignment: AgentGuard already speaks that standard, so the scanner understands the skill format you load. It also already scans OpenClaw skills, which is most of your skill surface.
Intelligence-Per-Watt alignment: the scan adds negligible FLOPs / energy / latency compared to a single inference call, so it fits the project's first-class constraints.

[...]

PR #601, opened 2026-06-27
feat: add optional before_tool_call governance hook to OrchestratorAgent

foundationx:
Adds an optional before_tool_call: Callable[[str, dict], bool] param to OrchestratorAgent. The hook fires before each tool executes across all three execution paths (sequential function-calling, parallel function-calling, and structured ReAct mode). Returning False injects a denial message instead of running the tool; returning True allows normal execution. Hook defaults to None -- no behaviour change for existing users.

Adds TestOrchestratorGovernanceHook to tests/agents/test_orchestrator.py with 6 tests covering every execution path and both allow/deny outcomes. 35/35 tests pass. Ruff lint clean; ruff format applied to changed files.

## What does this PR do?

Adds an optional before_tool_call governance hook to OrchestratorAgent so callers can intercept and approve or deny tool calls before execution. The hook is wired into all three execution paths (sequential, parallel, and structured ReAct). Existing users see no behaviour change.

## How was this tested?

Added TestOrchestratorGovernanceHook (6 tests) to tests/agents/test_orchestrator.py covering: no hook set, hook returns True, hook returns False, correct name/args delivered to hook, denial in parallel mode, denial in structured mode. All 35 tests pass.

## Checklist

- [X] Tests pass (`uv run pytest tests/ -v`)
- [X] Linter passes (`uv run ruff check src/ tests/`)
- [X] Formatter passes (`uv run ruff format --check src/ tests/`)
- [X] New/changed public API has docstrings


ElliotSlusky:
Thanks for the PR. Are you trying to complete #472? This is directionally useful, and I’d like to keep it open, but I don’t think it should merge as-is.

A few things need tightening first:
- If `before_tool_call` raises, the agent currently crashes. Since this is a governance hook, please catch hook errors and fail closed with a denied `ToolResult`.
- Invalid JSON currently gets passed to the hook as `{}`, while the executor would reject it. Please preserve enough raw argument context for policy review, or run the hook after the same argument parsing behavior used by the executor.
- In parallel tool mode, the hook runs from worker threads. Please either document that callbacks must be thread-safe or evaluate the hook before dispatching parallel work.

ElliotSlusky:
Maintainer repair pushed in `088923b` and integrated with current `main` in `0d5270a`:

- governance-hook exceptions now fail closed with a denial result instead of aborting the agent
- malformed/non-object arguments fail closed and never reach the documented dict callback
- policy callbacks are evaluated deterministically on the request thread before approved tools are dispatched in parallel, so stateful hooks are not invoked concurrently
- one failed parallel decision no longer prevents other approved calls from running

Validation on current main: **78 passed** across orchestrator and structured-orchestrator coverage; Ruff check/format passed.

ElliotSlusky (approved):
Maintainer repair reviewed on current main: 78 focused tests and Ruff pass. Hook errors fail closed and callback invocation is serialized before parallel tool execution.

foundationx:
I am sorry about the silence here. I genuinely do not know why I never got a notification
for your review. I searched my email and there is nothing from this thread until last
night, seven weeks after you posted it, which is when it finally showed up and I saw it
for the first time. I am still trying to work out what went wrong on my end. Either way
you had every reason to read this as abandoned, and I am sorry you had to wait on it.

Thank you for doing the repair yourself rather than closing it out. I went through
`088923b`. Two things in it went past what you had originally asked me for: denying
non-dict arguments that parse as valid JSON, and stopping one failed parallel decision
from blocking the other approved calls. Both are better than what I would have written.
Denying outright on invalid JSON, rather than preserving raw context for the callback, is
also the right call, since it keeps a single parse path that cannot drift from the
executor.

To answer the question you asked in that review: no, I was not working toward #472. I was
not aware of that issue when I opened this, and only read it this week. I have posted a
note there covering what the merged hook does and does not cover, so the overlap is on the
record.

Issue #608, opened 2026-06-29
estimate_prompt_tokens undercounts tool-call turns (tool_calls args + reasoning_content not counted)


gilbert-barajas:
Follow-up from #276. The crash fix counts a `None`-content message as 0 chars — that stops the `TypeError`, but it means a tool-call assistant turn now contributes nothing to the prompt-token estimate, even though its `tool_calls` arguments (often the bulk of the turn) and any `reasoning_content` are real tokens on the next request. So `estimate_prompt_tokens` returns a confidently-low number for tool-heavy conversations, which feeds the cost / FLOPs / energy math downstream.

Two parts:
1. Count `tool_calls` arguments (and `reasoning_content` where present) in `estimate_prompt_tokens`, not just `message.content`.
2. `Message.content` is annotated `str` (`core/types.py`) but legitimately holds `None` on tool-call turns — the annotation should be `str | None` (and mypy in CI would have caught the original #276 bug). A small `Message.text` property coercing `None`->`""` would also centralize the guard currently duplicated across call sites.

Not urgent — the crash is resolved by #276 — but worth tracking so the estimate doesn't silently drift on tool-using agents.

Issue #607, opened 2026-06-29
Bug: native_openhands token estimate has the same None-content len() crash as #276


gilbert-barajas:
While fixing #276 (`estimate_prompt_tokens` crashing on a tool-call turn's `content=None`), a grep turned up an identical, independent copy of the pattern in the OpenHands agent:

`src/openjarvis/agents/native_openhands.py:119`
```python
total_chars = sum(len(m.content) for m in messages)
```

Same failure mode as #276: a tool-call assistant turn legitimately carries `content=None` (OpenAI chat schema; Qwen3 on Ollama produces it), so the next turn's estimate raises `TypeError: object of type 'NoneType' has no len()`. The #276 fix lives in `engine/_base.py` and doesn't touch this copy, so the OpenHands agent path stays exposed.

Deterministic repro:
```python
from openjarvis.core.types import Message, Role
msgs = [Message(role=Role.USER, content="hi"), Message(role=Role.ASSISTANT, content=None)]
sum(len(m.content) for m in msgs)  # TypeError: object of type 'NoneType' has no len()
```

Suggested fix: the same guard (`len(m.content or "")`), or factor the shared estimate into one helper so both call sites stay in sync.

PR #609, opened 2026-06-29
fix(engine): guard None content in estimate_prompt_tokens (#276)

### LINKED ISSUE #276: [Bug] Ollama + Qwen 3-Coder + Frontend agents messaging fails with server side error
### Description

I just installed OpenJarvis globally and I am trying to set up the frontend web browser app. The model I'm using is Qwen 3-Coder with Ollama 0.21.0 and I'm actually on **Python 3.14** not 3.13. (This form menu only goes to 3.13.) I'm trying to set up an agent with the Agents dashboard. When I finish creating it (manual trigger) and click "run now", nothing seems to happen until I send it a generic message. Then after the first or a few turns, the agent usually silently fails, with the attached error in the server console log about the response (`m.content`) being `None`.

Having recently tried to make an agent with the Ollama Python API and Qwen, I strongly suspect this is something to do with the text response from the Qwen model actually appearing in `response.thinking` not `response.content` (an Ollama bug) but I will investigate and report back in this thread.

### Steps to Reproduce

1. Run `uv run jarvis serve --port 8000` in the server and `npm run dev` in the frontend
2. When the server is warmed up and connected, create an agent in the Agent dashboard with the "Code Reviewer" template (I only changed the system prompt/description, but it doesn't seem to matter what that is for this error)
3. Click "Launch Agent" -> nothing seems to happen
4. Send a message in the agent's chat
5. It might run for a bit and then stop without completing the task. The model is still loaded in memory. The indicator on the dashboard says the agent is still running until I click "Pause".

# Screenshots:
## Agent chat dashboard when the agent stops running and the error appears in the server console
Image

## Agent overview
Image

### Expected Behavior

The agent runs without error, outputs its thinking to the chat, and completes the task.

### Actual Behavior

The agent stops outputting or appearing to do anything in the dashboard, and the server error about the model response being `None` appears in the server console.

### Operating System

Windows

### Python Version

3.13

### Hardware

NVIDIA GPU

### Engine

Ollama

### Logs / Traceback

```shell
ERROR openjarvis.server.agent_manager: Managed agent stream error: object of type 'NoneType' has no len()
Traceback (most recent call last):
  File "C:\Users\the-m\Documents\Tools\OpenJarvis\src\openjarvis\server\agent_manager_routes.py", line 1050, in generate
    async for chunk in engine.stream_full(
    ...<34 lines>...
            current_finish_reason = chunk.finish_reason
  File "C:\Users\the-m\Documents\Tools\OpenJarvis\src\openjarvis\telemetry\instrumented_engine.py", line 486, in stream_full
    async for chunk in self._inner.stream_full(
    ...<6 lines>...
        yield chunk
  File "C:\Users\the-m\Documents\Tools\OpenJarvis\src\openjarvis\engine\ollama.py", line 275, in stream_full
    async for chunk in self._run_stream(
    ...<2 lines>...
        yield chunk
  File "C:\Users\the-m\Documents\Tools\OpenJarvis\src\openjarvis\engine\ollama.py", line 346, in _run_stream
    est_prompt = estimate_prompt_tokens(messages)
  File "C:\Users\the-m\Documents\Tools\OpenJarvis\src\openjarvis\engine\_base.py", line 54, in estimate_prompt_tokens
    total_chars = sum(len(m.content) for m in messages)
  File "C:\Users\the-m\Documents\Tools\OpenJarvis\src\openjarvis\engine\_base.py", line 54, in <genexpr>
    total_chars = sum(len(m.content) for m in messages)
                      ~~~^^^^^^^^^^^
TypeError: object of type 'NoneType' has no len()
```

gilbert-barajas:
## Summary

Agents on Ollama with Qwen3-Coder crash mid-conversation with `TypeError: object of type 'NoneType' has no len()`.

A tool-call assistant turn legitimately carries `content=None` (OpenAI chat schema), and Qwen3 on Ollama produces exactly this — text routes through the thinking channel, so `content` comes back empty and the assistant message is stored with `content=None`. On the **next** turn, `estimate_prompt_tokens` sums `len(m.content)` over the history with no None guard and raises.

Guard with `len(m.content or "")` — a None-content message contributes 0 chars, the correct estimate for a tool-call-only turn. One line; no change for existing string-content messages.

Closes #276.

## Testing

`tests/engine/test_base.py` — `estimate_prompt_tokens` had no coverage:
- a None-content assistant message no longer raises (exact token count asserted)
- a delta test pins that a None message adds only per-message overhead, so a future `len(m.content)` regression re-raises here
- a basic multi-message count

Reproduced the crash deterministically and against a real Qwen3 model on Ollama before the fix; green after. Full engine suite passes (439), `ruff check` + `ruff format --check` clean.

## Follow-ups

Found while fixing this, filed separately to keep the PR to one change:
- #607 — `native_openhands.py` has an identical `sum(len(m.content) ...)` (same crash, different agent)
- #608 — `estimate_prompt_tokens` undercounts tool-call turns (`tool_calls` args / `reasoning_content` uncounted); `Message.content` typed `str` but holds `None`

### DIFF
--- a/src/openjarvis/engine/_base.py
+++ b/src/openjarvis/engine/_base.py
@@ -55,7 +55,7 @@ def estimate_prompt_tokens(messages: Sequence[Message]) -> int:
     Uses ~4 characters per token (standard BPE average for English) plus
     a small per-message overhead for role markers and separators.
     """
-    total_chars = sum(len(m.content) for m in messages)
+    total_chars = sum(len(m.content or "") for m in messages)
     # ~4 tokens overhead per message for role markers / separators
     overhead = len(messages) * 4
     return max(1, total_chars // 4 + overhead)
--- a/tests/engine/test_base.py
+++ b/tests/engine/test_base.py
@@ -1,0 +1,36 @@
+"""Tests for openjarvis.engine._base utility helpers."""
+
+from __future__ import annotations
+
+from openjarvis.core.types import Message, Role
+from openjarvis.engine._base import estimate_prompt_tokens
+
+
+class TestEstimatePromptTokens:
+    def test_counts_basic_messages(self):
+        # 25 content chars // 4  +  2 messages * 4 overhead
+        msgs = [
+            Message(role=Role.USER, content="hello there"),
+            Message(role=Role.ASSISTANT, content="general kenobi"),
+        ]
+        assert estimate_prompt_tokens(msgs) == 14
+
+    def test_handles_none_content(self):
+        # A tool-call assistant turn legitimately carries content=None per the
+        # OpenAI chat schema; Qwen3 on Ollama produces exactly this shape. The
+        # next turn's prompt estimate must not crash. Regression for the
+        # "TypeError: object of type 'NoneType' has no len()" in issue #276.
+        msgs = [
+            Message(role=Role.USER, content="What is 17 * 23?"),  # 16 chars
+            Message(role=Role.ASSISTANT, content=None),
+        ]
+        # 16 // 4  +  2 messages * 4 overhead = 4 + 8
+        assert estimate_prompt_tokens(msgs) == 12
+
+    def test_none_message_adds_only_overhead(self):
+        # A None-content message must contribute its per-message overhead and
+        # zero content chars. Pins the fix so a future `len(m.content)`
+        # regression (which would re-raise the #276 TypeError) is caught.
+        base = [Message(role=Role.USER, content="What is 17 * 23?")]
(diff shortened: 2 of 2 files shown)


ElliotSlusky:
Thanks for this fix @gilbert-barajas. The same `estimate_prompt_tokens()` guard has now landed through #612, which also fixed the related native OpenHands path.

Since this PR now conflicts with `main` and the production change is already merged, I’m going to close this as superseded. Appreciate the clean repro and tests here. This was really useful!

PR #626, opened 2026-07-03
Make OpenAI-compat and Ollama engine streaming truly async (unblock the event loop)

talyaseen:
## Problem

The OpenAI-compatible and Ollama engines expose `stream()` / `stream_full()` as `async def`, but internally they iterate a **synchronous** `httpx.Client.iter_lines()`. Because the sync client blocks the thread on every inter-token read, each of those blocking reads stalls the single asyncio event loop:

- Concurrent chat requests **serialize** instead of interleaving — while one stream waits between tokens, no other request can make progress.
- A single wedged upstream read (slow or hung model server) **freezes the whole API**, not just the one request.
- The configured request `timeout` was never applied to the streaming reads, so a hung read could hang indefinitely.

## Fix

Convert both engines' streaming paths to a genuinely non-blocking implementation:

- **Per-call `httpx.AsyncClient` + `aiter_lines()`** — inter-token reads now yield to the event loop, so concurrent streams interleave and one slow read no longer blocks the others. The client is opened/closed by `async with`, keeping the async lifecycle self-contained.
- **Bounded read timeout** — the engine's configured `timeout` is carried into the async client, so a wedged read fails at the bound (`ReadTimeout`) instead of hanging.
- **Typed context-overflow error** — a new `EngineContextLengthError` (subclass of `EngineConnectionError`, so existing handlers still catch it) is raised when an upstream 400 body reads like a context-window overflow, letting callers show a clear "conversation too long" message. Ollama has no such signal, so it has no context branch.
- **Mid-stream transport errors mapped** — `httpx.RemoteProtocolError` / `httpx.ReadError` (server dying between tokens) are mapped to `EngineConnectionError` alongside `ConnectError` / `TimeoutException`. The except clause is kept deliberately narrow so `asyncio.CancelledError` / `GeneratorExit` (not `TransportError` subclasses) keep propagating for correct cancellation.
- **Non-2xx handled explicitly** — `not resp.is_success` covers 3xx as well as 4xx/5xx. With redirects off (the default) an unexpected 3xx previously fell through to `aiter_lines()` and surfaced as a **silent empty stream**; it now raises a clean engine error. The short error body is `aread()` before touching `.text`.

**No public API change** — signatures and return types are unchanged; only the internal transport becomes async.

## Tests

All new tests use `httpx.MockTransport` (no real server), covering:

- the async streaming path yields tokens correctly,
- the configured timeout is applied to the async client,
- a mid-stream disconnect maps to `EngineConnectionError`,
- HTTP-error mapping (incl. 3xx and context-length 400 -> `EngineContextLengthError`).


ElliotSlusky:
This is a really strong PR — thank you! The core diagnosis (sync `iter_lines()` inside `async def` stalling the whole event loop) is exactly right, the fail-narrow exception handling with cancellation-safety comments is genuinely careful work, and the `MockTransport`-based tests are a pleasure to read: the "hardened" timeout pins that verify the timeout reached the wire (not just the engine attribute) are the kind of tests that keep this fixed forever.

I reviewed it in depth, and since maintainer edits were enabled I pushed three commits on top rather than bouncing it back — I hope that's welcome; happy to talk through any of it. What they address:

**1. Restored connection pooling (`ace9cc71`).** The per-call `AsyncClient` made every conversation turn pay a fresh TCP+TLS handshake against remote hosts, where the old shared sync client had pooled keep-alive connections. `AsyncClient` is safe to share across coroutines, so the shared plumbing now caches one client per event loop (per-loop because pooled connections die with their loop — CLI flows that `asyncio.run()` per turn still work) and `close()` tears it down. Your `_async_transport` test seam and all existing tests carry over unchanged.

**2. Deduplicated the mirrored helpers (also `ace9cc71`).** `_make_async_client` / `_raise_stream_http_error` / the transport-error tuple were near-identical between the two engines (your docstrings even said "Mirrors…" 🙂). They now live in one `AsyncHTTPEngineMixin` in `engine/_http_async.py`, so the third engine that needs this (see below) gets it for free.

**3. Tightened the context-length markers (also `ace9cc71`).** Bare `"please reduce"` / `"too many tokens"` also match unrelated 400 bodies — e.g. `"please reduce max_tokens and retry"` or a rate-limit "Too many requests, please reduce your request rate" — which would show users "conversation too long, start a new chat" for errors that have nothing to do with conversation length. The markers are now all anchored on "context" (your vLLM-style test still passes via "maximum context length"), with a regression test for the false-positive case.

**4. Fixed the same bug in the LiteLLM engine (`8f21534b`).** `LiteLLMEngine.stream` had the identical pattern — `async def` calling the sync `litellm.completion(stream=True)` with a plain `for` loop — and it's reachable through the same server path. Switched to `acompletion` + `async for`.

**5. Wired up `EngineContextLengthError` and unblocked the non-streaming paths (`8a3bbcb5`).** The new error type was raised but nothing consumed it: `cli/ask.py` caught it as a base `EngineConnectionError` and told users to check whether their server was running, the agent retry classifier treated it as retryable (burning ~30s of backoff on an error that deterministically recurs), and `stream_bridge.py`'s own string check didn't match the new message's "context window" wording. All three now recognize it (via one shared `looks_like_context_length_error()` helper, replacing three divergent copies across the codebase). And since your PR description correctly called out that a wedged upstream "freezes the whole API": the *non-streaming* paths had the same bug — `routes.py` and the websocket fallbacks called sync `engine.generate()` / `agent.run()` directly on the event loop — now dispatched via `asyncio.to_thread`.

Test results: engine + server + agents suites green locally (495 passed; the only failures in the wider suite are the known pre-existing Rust-extension ones, unchanged by this branch).

Thanks again — this PR fixed a real operational pain point and the foundation it laid made the follow-ups easy to build on. 🚀

ElliotSlusky (approved):
Reviewed in depth (see comment above for the full rundown): the core async conversion is correct and well-tested, and the follow-up commits on this branch address everything the review surfaced — connection-pool reuse, the LiteLLM engine's identical blocking bug, consuming the new context-length error end-to-end, tightened error markers, and the non-streaming server paths. All 12 checks green. Thanks @talyaseen!

ElliotSlusky:
@talyaseen one last mechanical step to land this: the repo's branch rules require the *most recent push* to be approved by someone other than its pusher — and since the review follow-up commits came from me, my approval can't certify my own push. 🙂

Could you push any trivial update to your `fix/async-streaming` branch (an empty commit is fine: `git commit --allow-empty -m "chore: trigger review" && git push`, or just click **Update branch** on this page if it's offered)? The moment you do, I'll re-approve and merge — everything is already green (all 12 checks passing). Thanks!

ElliotSlusky (approved):
All checks green on the latest push — merging. Thanks again for the quick turnaround, @talyaseen!

talyaseen:
> All checks green on the latest push — merging. Thanks again for the quick turnaround, @talyaseen!

My Pleasure. I'm guessing the other code owner's need to review as well?

PR #629, opened 2026-07-07
fix(memory): make auto-extracted facts recallable at inference

finance180:
## Problem

Automatic memory is effectively **write-only**: facts extracted during `jarvis chat`/`serve` land in the JSONL fact store (visible via `jarvis memory list`) but are never surfaced by `jarvis ask`/`chat`. Two independent gaps cause this.

### 1. No fact→retrieval bridge

Fact extraction writes to the `FactStore` (JSONL), but inference-time context injection (`inject_context` / the executor) reads a **separate** FTS document store (`sqlite`/`bm25`/…). Nothing ever indexed extracted facts into that store, so recall always queried an empty index.

`MemoryService` now takes an optional `retrieval_backend` and mirrors each **newly stored** fact into it:
- dedup-aware — only facts that are genuinely new (`FactStore.add()` returns `True`) get indexed, so restarts don't accumulate duplicate documents;
- best-effort — indexing failures are swallowed so a flaky retrieval backend can neither drop a fact nor kill the extraction worker.

`build_memory_service` constructs the backend from config (mirroring `cli.ask._get_memory_backend`) and wires it in, so the bridge is active by default when memory is enabled.

### 2. FTS5 query silently died on punctuation

The Rust `sqlite` backend only trimmed *leading/trailing* punctuation from each term, so an **interior** apostrophe (`"dog's"`) reached the FTS5 `MATCH` as a syntax error — which row iteration then swallowed via `.filter_map(|r| r.ok())`, yielding **zero results with no error**. So "What is my dog's name?" retrieved nothing.

Each term is now wrapped as a quoted FTS5 phrase (embedded double-quotes doubled per the FTS5 escaping rule); the tokenizer ignores the punctuation inside the phrase. Queries that reduce to only punctuation now return an empty result set instead of erroring.

## Testing

- **Rust** (`storage::sqlite::tests`): new `test_sqlite_apostrophe_in_query` and `test_sqlite_punctuation_only_query_is_empty_not_error`. Full crate: 47 passed.
- **Python** (`tests/memory/test_memory_service.py`): index-new-facts, no-duplicate-reindex, best-effort-on-backend-failure, backward-compat-without-a-backend, and build-wires-backend. Memory + related CLI suites: 167 passed.
- `cargo clippy --all-targets -- -D warnings` clean; `ruff check`/`format` clean.

## Manual verification (local MLX + Qwen2.5-7B-Instruct)

1. Fresh `jarvis chat`: "My dog is named Biscuit and my favorite color is orange." → `memory list` shows 2 facts **and** `memory stats` shows 2 retrieval documents — indexed automatically, no manual `memory index` step.
2. Fresh `jarvis ask "What is my dog's name?"` (note the apostrophe) → **"Your dog's name is Biscuit."** Both previously failed.

### DIFF
--- a/rust/crates/openjarvis-tools/src/storage/sqlite.rs
+++ b/rust/crates/openjarvis-tools/src/storage/sqlite.rs
@@ -144,16 +144,21 @@ impl MemoryBackend for SQLiteMemory {
     ) -> Result<Vec<RetrievalResult>, OpenJarvisError> {
         let conn = self.conn.lock();
 
+        // Wrap every term as a quoted FTS5 string literal. FTS5 metacharacters
+        // (apostrophes, quotes, parens, etc.) are otherwise interpreted as
+        // query syntax — an interior apostrophe like "dog's" is a syntax error
+        // that iteration silently swallows into zero results. Quoting makes each
+        // term a phrase whose tokenizer ignores the punctuation, so natural-
+        // language questions match. Embedded double quotes are doubled per the
+        // FTS5 escaping rule; terms that tokenize to nothing are dropped.
         let words: Vec<String> = query
             .split_whitespace()
-            .map(|w| w.trim_matches(|c: char| "?.,!;:'\"()[]{}/ ".contains(c)).to_string())
-            .filter(|w| !w.is_empty())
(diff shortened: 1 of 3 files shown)


finance180:
Closing — opened in error; this was intended as a local change only.

PR #630, opened 2026-07-08
Fix proactive-agent cron duplicating on every server restart

esslamali15:
## Problem
`register_cron()` runs on every `jarvis serve` startup and `create_task()` persists to scheduler.db — so **every restart adds one more copy** of the daily proactive cron. On my install 68 duplicates had accumulated; when the schedule fired they all ran back-to-back and monopolized the (single-slot) Ollama queue, stalling interactive Telegram chats behind minutes of duplicate proactive runs.

## Fix
- `register_cron()` is now idempotent: it looks up an existing active task with the same agent+prompt, reuses it, and cancels surplus duplicates left over from earlier restarts (self-healing for already-affected installs).
- The notification channel is constructed without calling `connect()`: it only needs to *send*, and starting a second getUpdates poll loop with the same bot token causes a Telegram `Conflict` error that kills the main channel's listener.

## Testing
Verified on a live install: after restart, `scheduler.db` shows exactly 1 active proactive task (was 68), and Telegram polling stays alive with proactive notifications enabled.

### DIFF
--- a/src/openjarvis/agents/proactive_agent.py
+++ b/src/openjarvis/agents/proactive_agent.py
@@ -37,6 +37,7 @@ called from your app startup:
 from __future__ import annotations
 
 import json
+import logging
 from datetime import datetime
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Set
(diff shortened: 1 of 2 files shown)


ElliotSlusky (approved):
Reviewed the updated cron reconciliation, channel lifecycle handling, notification destination parsing, and regression coverage.

PR #631, opened 2026-07-08
browser_click: accept param aliases and resolve natural-language descriptions

esslamali15:
## Problem
Local models (qwen3, llama, etc.) driving the orchestrator call `browser_click` with `target=`/`text=`/`element=` instead of `selector=`, and pass descriptions like *'first video result'* rather than CSS. Both dead-end today (`No selector provided` / Playwright invalid-selector error), which in practice breaks most browse-and-click workflows — the model gets an unhelpful error, retries the same thing, and eventually apologizes to the user.

## Fix
- **Param aliases**: `selector` | `target` | `text` | `element` | `query` all accepted.
- **Strategy ladder**: CSS → exact text → fuzzy text → YouTube video-title special-case → role-based keyword match (with ordinal support: *first/second/last*). Existence is probed with `count()` so non-matching rungs cost milliseconds; only a matching strategy gets a click timeout.
- **Iframe search**: after the main frame, child frames are searched (cookie-consent dialogs and ad overlays live there).
- **Self-correcting errors**: on failure the error lists the visible clickable elements, so the model's next attempt has real candidates to choose from.

## Testing
On a live install (qwen3:8b orchestrator): the previously-failing flow *navigate to YouTube search → click 'first video result'* now resolves in ~0.1s via the ladder and lands on the correct watch page; a nonsense selector returns the candidate list instead of a bare error.


ElliotSlusky:
Maintainer repair pushed in 4001161d after merging current main append-only.

The resolver now examines every matching locator and skips hidden/disabled duplicates instead of repeatedly choosing .first. Explicit ordinals remain deterministic, while the click itself remains Playwright's final actionability check. The accepted target/text/element/query aliases are now represented in the JSON schema, so alias calls are not invalidated by a selector-only required field.

Validation:
- PYTHONPATH=src python3 -m pytest tests/tools/test_browser.py -q: 73 passed
- ruff on changed browser code/tests: passed

ElliotSlusky:
CI follow-up: repository-wide ruff format --check identified browser.py after the first push. Pushed formatting-only commit 87a72922 after rechecking the live head (non-force). Focused browser tests remain 73 passed, and local ruff format --check now passes.

ElliotSlusky (dismissed):

PR #632, opened 2026-07-10
feat(memory): make auto-fact trust tag load-bearing at recall

TinkerOfThings:
Follow-up to #599, resolving Elliot's review point that `trust="untrusted"` is inert at inference time because `LocalFactStore` is disjoint from the model-facing recall path.

### The gap
Auto-extracted facts are captured from raw conversation text that may contain untrusted target output (scraped pages, firmware strings, recon dumps). #599 tags them `trust="untrusted"` in the `FactStore` — but nothing reads that tag at inference time. `LocalFactStore` is only read by the `memory list` CLI; the model-facing `MemoryBackend` recall path never sees it. So the tag was decorative.

### The fix — two pieces
1. **Mirror.** `MemoryService` now mirrors each *newly-stored* auto fact into the model-facing `MemoryBackend` via `store(text, source="auto", metadata={"trust": "untrusted"})`. Best-effort: an absent, unavailable, or raising backend never takes down extraction. Only the persistent shared **sqlite** backend closes the loop end-to-end; in-memory per-instance backends (bm25) are a documented limitation, not a silent failure.
2. **Filter.** `apply_trust_policy`, wired into `inject_context`, reads `metadata["trust"]` and — per `ContextConfig.untrusted_policy` — **drops** untrusted results (default) or **annotates** them with an unverified-origin warning before they reach the model. Exposed as `memory.context_untrusted_policy` across all seven `ContextConfig` sites and the config echo.

### Tests prove the security property end-to-end
- An untrusted fact does **not** surface at recall by default (drop).
- A trusted one still does.
- The annotate path surfaces untrusted content **only** behind the caveat.
- Includes a live `mirror → recall` integration test over the real sqlite backend.

`161 passed, 14 skipped` across `tests/memory`.


ElliotSlusky:
Maintainer repair pushed in ce3cca09 (append-only merge of current main; no force-push).

Trust filtering now fails closed for unknown policy values, unknown trust labels, and malformed metadata while preserving legacy untagged documents. It is also applied to the managed-agent direct recall path. The main-branch durable-fact budget behavior is retained, including the case where all retrieved documents are filtered but facts remain. Owned mirror backends and ephemeral CLI/SDK/chat backends are closed without closing borrowed backends.

Validation:
- focused memory, CLI chat, and SDK tests: 82 passed, 4 environment-dependent async/sqlite skips
- ruff on changed memory/recall/CLI/SDK code and tests: passed

ElliotSlusky (approved):

ElliotSlusky:
Closing as superseded at the design level. Merged #740 made the JSONL fact store the direct recall source specifically to avoid duplicate copies in memory.db; this branch mirrors automatic facts back into memory.db and therefore conflicts with the current source-of-truth design. The valid trust/quarantine requirement should be implemented directly in #599 on the JSONL recall path, including safe promotion and locked mutation. This does not dismiss the underlying security concern. The contributor branch is retained.

TinkerOfThings:
Agreed — this closure is correct, and thanks for the detailed reasoning.

The mirror half of this branch does conflict with the current design. #740 made the JSONL fact store the direct recall source specifically to avoid duplicate copies in `memory.db`, and mirroring auto facts back into `memory.db` cuts against that. And the trust/quarantine requirement for facts has since landed on the JSONL path in #599 — I checked `main` and it's all there: `Fact.trusted_for_recall` failing closed on unknown tiers, `_trusted_facts()` filtering in both `inject_context` and `build_context_message`, `load_configured_facts()` filtering for the ask/SDK/orchestrator/chat paths while `FactStore.list()` retains quarantined rows for audit, and the collision rule keeping the least-trusted label on dedupe. That's the security property I was after, and it's enforced in a better place than where I put it. No objection to the close.

One narrow piece isn't covered by any of that, so I've opened it separately as #882.

`apply_trust_policy` in this branch operated on `RetrievalResult.metadata` — the **document** backend, not just facts. `main` has no equivalent: retrieved documents travel the same path into the same prompt as facts, but have no trust concept at all. #882 adds only that, reusing `RECALLABLE_TRUST_TIERS` rather than a parallel vocabulary, with no mirror, no source-of-truth change, and no new config knob.

Worth saying plainly: **it's inert today.** Nothing currently writes trust metadata into the document backend, so untagged documents behave exactly as before and no user-visible behaviour changes. It closes a fail-open gap ahead of anything starting to tag documents rather than fixing a live exploit — I'd rather state that than repeat the overclaiming you correctly caught on #599.

Happy for this branch to stay closed.

PR #636, opened 2026-07-13
fix: catch ImportError in git tools, fall back to CLI when Rust ext missing

Curryraj:
## Summary
- `get_rust_module()` was called *outside* the `try` block in `GitStatusTool`/`GitDiffTool`/`GitLogTool.execute()` (`src/openjarvis/tools/git_tool.py`), so on machines without the compiled `openjarvis-rust` extension the resulting `ImportError` bypassed the `except` handler entirely and propagated as an uncaught exception.
- Downstream, this surfaced through `ToolExecutor`'s generic exception handler as a vague "Tool execution error: ...", which caused a local model to hallucinate a fake result instead of reporting the real failure.
- Fix: move the `get_rust_module()` call inside `try`, and fall back to `subprocess` git (via the existing `_run_git` helper) on `ImportError` for all three tools — matching the fallback pattern `git_log` already partially had. `git_commit` was already CLI-only, unaffected.

## Test plan
- [x] `git_status`, `git_diff`, `git_log` all verified locally on a machine without the Rust extension built — each returns `success=True` with real git output via the CLI fallback.
- [x] `python -m py_compile` on the changed file.

### DIFF
--- a/src/openjarvis/tools/git_tool.py
+++ b/src/openjarvis/tools/git_tool.py
@@ -139,8 +139,8 @@ class GitStatusTool(BaseTool):
 
     def execute(self, **params: Any) -> ToolResult:
         repo_path = params.get("repo_path", ".")
-        _rust = get_rust_module()
         try:
+            _rust = get_rust_module()
             output = _rust.GitStatusTool().execute(repo_path)
             return ToolResult(
                 tool_name="git_status",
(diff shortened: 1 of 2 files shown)


ElliotSlusky (approved):
Verified live in a venv without the compiled extension: base code raises an uncaught ImportError from all three read-only git tools; with this fix they return real git output via the CLI fallback, and a non-repo path fails gracefully. Pushed a follow-up commit adding regression tests for the fallback path (the existing tests all mock a working Rust module, so this path had no coverage). 45/45 tests pass.

Issue #540, opened 2026-06-13
Odd references to claude / Identity crisis


cpass78:
### Description

latest version 1.0.2, see screenshot 

### Steps to Reproduce

<img width="707" height="864" alt="Image" src="https://github.com/user-attachments/assets/284efff2-e73b-4b13-8206-d0084d69a190" />

### Expected Behavior

To know what it is

### Actual Behavior

same as above

### Operating System

Linux

### Python Version

3.10

### Hardware

NVIDIA GPU

### Engine

Ollama

### Logs / Traceback

```shell

```

jonsaadfalcon:
Thanks again for the clear report + screenshot, @cpass78 — that made this easy to pin down. Fixed in #546: the desktop chat endpoint (`/v1/chat/completions`) now injects OpenJarvis's identity system prompt when the client doesn't send one, and the wording was hardened so models stop claiming to be Claude/etc. Verified locally — a model that answered "I'm Claude" now answers "I'm OpenJarvis." It'll be in the next release. 🙏

Issue #431, opened 2026-05-29
Custom agent streaming chat does not load SOUL.md / USER.md prompt context


selena731:
### Description

Custom agents using the streaming chat path appear to bypass SystemPromptBuilder, so SOUL.md, MEMORY.md, and USER.md are not injected into the system prompt

### Steps to Reproduce

1. Create custom SOUL.md, ex. "You are Jessica."
2. Create a Custom Agent in the browser dashboard
3. Send a "who are you" message in the chat to the agent

### Expected Behavior

Chat reply "I am Jessica.." 

I.e. custom agent chat should consistently load the same SOUL.md, MEMORY.md, and USER.md context as the run path.

### Actual Behavior

Chat reply "I am an AI language model created by OpenAI, designed to assist with a wide range of tasks by providing information, answering questions, and performing various functions. How can I assist you today?"

### Operating System

Windows

### Python Version

3.10

### Hardware

CPU only

### Engine

Cloud (OpenAI/Anthropic/Google)

### Logs / Traceback

```shell

```

SoulSniper-V2:
take

PR #637, opened 2026-07-13
fix: wire persona files into default chat endpoint, fix FTS5 apostrophe crash

tR1L3Yw:
## What does this PR do?

Two small, related fixes found while wiring OpenJarvis into a personal memory-vault setup:

1. **`/v1/chat/completions` now applies SOUL.md/MEMORY.md/USER.md persona files.** `_ensure_identity_prompt` in `routes.py` previously only injected the bare `agent.default_system_prompt` blurb (added for #540) and never went through `SystemPromptBuilder`. That meant persona files applied to `jarvis ask`, managed agents, and persistent agents, but silently *not* to the main dashboard/OpenAI-compatible chat endpoint — the primary path most UIs actually use. This now builds the full persona-aware prompt the same way `agent_manager_routes.py` already does (see its `_build_managed_system_prompt`, added for #431), so behavior is consistent across all entry points.

2. **FTS5 search silently returned zero results for any query with an internal apostrophe.** In `sqlite.rs`, `retrieve()` tokenized by whitespace and only trimmed punctuation from word *edges*, so a token like `user's` was passed as-is into the FTS5 `MATCH` string. FTS5's query grammar treats an unescaped `'` as a string delimiter, so this silently fails to parse and returns zero rows — no error surfaced anywhere. Natural-language queries like "what's the user's name" would just come back empty. Fixed by splitting on any non-alphanumeric character instead of only whitespace, so punctuation never reaches the MATCH string.

Both were found via hands-on testing: seeding `SOUL.md` with a custom persona and confirming it worked through `jarvis ask` but not through the dashboard chat endpoint, and separately noticing `jarvis memory search "what is the user's name"` returned nothing despite the document being indexed and directly matchable via `jarvis memory search "Trev"`.

## How was this tested?

- Added `test_sqlite_apostrophe_in_query` (Rust, `cargo test -p openjarvis-tools`) covering both the multi-word OR-join path and the single-word bypass path that previously skipped it entirely.
- Added `test_retrieve_query_with_apostrophe` (Python, `tests/memory/test_sqlite.py`) as an equivalent regression test at the Python API boundary.
- Added `test_direct_injects_soul_persona_when_present` (Python, `tests/server/test_routes.py`) confirming `/v1/chat/completions` now injects SOUL.md content alongside the identity blurb, following the existing `TestIdentityPromptInjection` conventions.
- Ran the full suite (`uv run pytest tests/ -n auto -q -m "not live and not cloud and not hub"`): 7154 passed, 12 failed, 7 errored. All failures are pre-existing and unrelated to this change — CLI/ANSI-output assertion brittleness (`test_agent_cmd`, `test_config_cmd`, `test_doctor_bg`, `test_digest_schedule`, `test_memory_cmd`'s clear command, `test_quickstart`) and the dense/FAISS embedding backend requiring a live `/api/embed`-capable Ollama endpoint not configured in this environment (`test_dense.py`), plus one unrelated `evals/` import error. None touch `routes.py`, `sqlite.rs`, or the two test files changed here.
- Ran `cargo clippy -p openjarvis-tools --all-targets -- -D warnings` (clean, matches CI's gate).
- Ran `ruff check` / `ruff format --check` on all changed Python files (clean).

## Checklist

- [x] Tests pass (`uv run pytest tests/ -v`) — full suite run above; no failures related to this change
- [x] New/changed public API has docstrings — n/a, no public API surface changed
- [ ] Follows registry pattern (if adding new component) — n/a, no new component
- [ ] Documentation updated (if applicable) — n/a, internal bug fixes only

### DIFF
--- a/rust/crates/openjarvis-tools/src/storage/sqlite.rs
+++ b/rust/crates/openjarvis-tools/src/storage/sqlite.rs
@@ -144,16 +144,21 @@ impl MemoryBackend for SQLiteMemory {
     ) -> Result<Vec<RetrievalResult>, OpenJarvisError> {
         let conn = self.conn.lock();
 
+        // Split on any non-alphanumeric character (not just whitespace) so
+        // internal punctuation — apostrophes in particular ("user's") — never
+        // reaches the FTS5 MATCH string. FTS5's query grammar treats an
+        // unescaped `'` as a string delimiter, so passing a raw token like
+        // `user's` through silently fails to parse and yields zero rows with
+        // no visible error. Splitting fully avoids needing to escape anything.
         let words: Vec<String> = query
-            .split_whitespace()
-            .map(|w| w.trim_matches(|c: char| "?.,!;:'\"()[]{}/ ".contains(c)).to_string())
+            .split(|c: char| !c.is_alphanumeric())
+            .map(|w| w.to_string())
             .filter(|w| !w.is_empty())
             .collect();
-        let fts_query = if words.len() == 1 {
-            words[0].clone()
-        } else {
-            words.join(" OR ")
-        };
+        if words.is_empty() {
+            return Ok(Vec::new());
+        }
+        let fts_query = words.join(" OR ");
 
         let mut stmt = conn
             .prepare(
--- a/rust/crates/openjarvis-tools/src/storage/sqlite.rs
+++ b/rust/crates/openjarvis-tools/src/storage/sqlite.rs
@@ -326,6 +326,27 @@ mod tests {
     }
 
     #[test]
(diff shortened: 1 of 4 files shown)


ElliotSlusky (approved):
Verified both fixes: the /v1/chat/completions prompt path now matches _build_managed_system_prompt exactly, and all 46 openjarvis-tools cargo tests pass including the new apostrophe regression test (which fails on the base tokenization, confirming it has teeth). Pushed a docstring touch-up so _ensure_identity_prompt's docs describe the SystemPromptBuilder resolution.

PR #638, opened 2026-07-13
fix(speech): close temp file before transcribing to fix Windows EACCES in faster-whisper backend

ScottyAmr:
## Summary

`FasterWhisperBackend.transcribe()` wrote the incoming audio bytes to a `tempfile.NamedTemporaryFile(suffix=..., delete=True)` and passed `tmp.name` to `model.transcribe()` **while the temp file was still open**.

On Windows this always fails: `NamedTemporaryFile` holds the file open with an exclusive handle (and `O_TEMPORARY`), so when faster-whisper hands the path to PyAV, PyAV cannot reopen it and raises:

```
av.error.PermissionError: [Errno 13] Permission denied: 'C:\\Users\\...\\Temp\\tmpXXXX.wav'
```

The result is that local speech-to-text via the `faster-whisper` backend is completely broken on Windows (it works on Linux/macOS, where a second open of an already-open file is allowed).

## Fix

- Create the temp file with `delete=False` and **close it before** calling `model.transcribe()`, so PyAV can reopen the path on Windows.
- Unlink the temp file in a `finally` block so it is cleaned up on both success and failure.

## Tests

Added two regression tests in `tests/speech/test_faster_whisper.py`:

- `test_faster_whisper_transcribe_temp_file_reopenable_and_removed` â€” the mocked `model.transcribe` reopens the temp path and reads the audio back, which reproduces the exact `[Errno 13]` failure on Windows against the old code (verified locally on Windows 11: fails before the fix, passes after). Also asserts the temp file is removed afterwards.
- `test_faster_whisper_transcribe_removes_temp_file_on_error` â€” the temp file is cleaned up even when transcription raises.

All 9 tests in `tests/speech/test_faster_whisper.py` pass.

ðŸ¤– Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/openjarvis/speech/faster_whisper.py
+++ b/src/openjarvis/speech/faster_whisper.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+import os
 import tempfile
 from typing import List, Optional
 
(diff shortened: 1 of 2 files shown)


ElliotSlusky (approved):
Verified: temp file is closed before model.transcribe() reopens it, segments iterator is fully consumed before the unlink, and all 9 speech backend tests pass. Pushed a small follow-up: the write is wrapped in the file's context manager so the handle closes even if write() raises, and the swallowed unlink OSError is now logged at debug.

PR #639, opened 2026-07-13
Enforce existing skill capability/trust-tier checks at install and ru…

jaaaaorr06:
…n time

The skills/security.py module already defined TrustTier classification and dangerous-capability detection (shell:execute, network:listen, filesystem:write), but nothing in the codebase actually called it - skills were imported and executed without any capability enforcement.

This wires the existing checks into the two places that matter:

- SkillImporter.import_skill(): classifies trust tier and checks for dangerous capabilities before writing anything to disk. Unreviewed skills (the default for anything pulled from Hermes, OpenClaw, or an arbitrary GitHub repo) that request dangerous capabilities are now refused unless the caller passes confirm_dangerous=True. Trust tier and dangerous capabilities are persisted to the .source metadata file.

- SkillExecutor.run(): re-validates the manifest's required_capabilities against an explicit allowed_capabilities set before running any step, independent of what happened at install time. Defaults to empty (deny dangerous-by-default) rather than trusting the manifest.

No behavior changes for skills that don't request dangerous capabilities - this only adds friction for skills that ask for shell/network/filesystem access from an unreviewed source.

## Checklist

- [ ] Linter passes (`uv run ruff check src/ tests/`)

### DIFF
--- a/src/openjarvis/cli/skill_cmd.py
+++ b/src/openjarvis/cli/skill_cmd.py
@@ -199,7 +199,16 @@ def _get_resolver(source: str, url: str = ""):
     default="",
     help="Repo URL (required when source is 'github').",
 )
-def install(query: str, with_scripts: bool, force: bool, url: str):
+@click.option(
+    "--yes-dangerous",
+    is_flag=True,
+    default=False,
+    help=(
+        "Confirm installing an unreviewed skill that requests dangerous "
+        "capabilities (shell/network-listen/filesystem-write)."
+    ),
+)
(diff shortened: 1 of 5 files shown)


ElliotSlusky (approved):
Reviewed and fixed up: runtime capability enforcement is now opt-in (None = no policy, so existing manager.py call sites keep their behavior; pass a set to enforce), skill install/sync gained the --yes-dangerous flag the refusal message referenced, sync surfaces refused skills, dead trust-tier branches removed, and both enforcement points now have test coverage (276 skills tests pass).

PR #640, opened 2026-07-14
fix(knowledge_sql): match write keywords on word boundaries (allow valid SELECTs)

Osamaali313:
## Problem

`KnowledgeSQLTool` in `src/openjarvis/tools/knowledge_sql.py` guards its read-only SQL path by rejecting any query that *contains* a write keyword as a bare substring:

```python
_FORBIDDEN = ("DROP", "DELETE", "INSERT", "UPDATE", "ALTER", "CREATE", "ATTACH")
for forbidden in _FORBIDDEN:
    if forbidden in normalized:   # substring, not word boundary
        return ToolResult(..., success=False)
```

So a perfectly valid read-only `SELECT` is rejected whenever a column, alias, or literal merely contains one of those words as a substring. This isn't hypothetical — the `knowledge_chunks` table has `deleted_at` and `created_at` columns (`connectors/store.py`), and the store's own retrieval filters with `WHERE deleted_at IS NULL`:

| valid read-only query | current | fixed |
|---|---|---|
| `SELECT content FROM knowledge_chunks WHERE deleted_at IS NULL` | ❌ "forbidden: DELETE" | ✅ runs |
| `SELECT author, created_at FROM knowledge_chunks ORDER BY created_at DESC` | ❌ "forbidden: CREATE" | ✅ runs |
| `SELECT author AS created_author FROM knowledge_chunks` | ❌ "forbidden: CREATE" | ✅ runs |

Real write statements stay blocked either way (the `startswith("SELECT")` gate catches a leading `DROP`/`DELETE`; a mid-query `; DELETE` still matches on a word boundary).

## Fix

Match write keywords on word boundaries with a compiled regex — exactly what the sibling tool `db_query.py` already does (`_WRITE_KEYWORDS = re.compile(...)`):

```python
_FORBIDDEN_RE = re.compile(
    r"\b(DROP|DELETE|INSERT|UPDATE|ALTER|CREATE|TRUNCATE|ATTACH)\b",
    re.IGNORECASE,
)
...
forbidden = _FORBIDDEN_RE.search(query)
if forbidden:
    return ToolResult(..., content=f"Query contains forbidden keyword: {forbidden.group(1).upper()}. ...", success=False)
```

## Test

Added `test_allows_select_with_keyword_substring` (a `created_author` alias contains `CREATE`). It fails on the current code (`forbidden keyword: CREATE`) and passes after the fix. Full `tests/tools/test_knowledge_sql.py` stays green (8 passed) — the existing `test_rejects_non_select` / `test_rejects_drop` cases (real writes) still pass.

### DIFF
--- a/src/openjarvis/tools/knowledge_sql.py
+++ b/src/openjarvis/tools/knowledge_sql.py
@@ -6,6 +6,7 @@ and filtering operations that BM25 search cannot handle.
 
 from __future__ import annotations
 
+import re
 import sqlite3
 from typing import Any, Optional
 
@@ -15,6 +16,14 @@ from openjarvis.core.types import ToolResult
 from openjarvis.tools._stubs import BaseTool, ToolSpec
 
 _MAX_ROWS = 50
+
+# Write keywords are matched on word boundaries (mirroring db_query.py) so that
+# a read-only SELECT is not rejected just because a column/alias/literal happens
+# to contain one as a substring (e.g. "deleted_at", "created_at").
+_FORBIDDEN_RE = re.compile(
+    r"\b(DROP|DELETE|INSERT|UPDATE|ALTER|CREATE|TRUNCATE|ATTACH)\b",
+    re.IGNORECASE,
+)
 
 _SCHEMA_DESCRIPTION = (
     "Table: knowledge_chunks\n"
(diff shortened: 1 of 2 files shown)


ElliotSlusky (approved):
Verified the fix: regression test fails on base and passes with the change; word-boundary matching cannot admit a real write (startswith-SELECT gate + sqlite3 single-statement enforcement). Pushed a small follow-up commit: string literals are stripped before the keyword scan, sqlite3.Error is caught instead of only OperationalError, and created_at/deleted_at are documented in the schema description.

PR #646, opened 2026-07-17
Fix(frontend): remove jitter when scrolling up during chat autoscroll

goatoush:
## What does this PR do?

Originally while text was streaming in the chat area in the desktop app it was hard to scroll up creating a jitter when you tried to scroll up, because interface activated autoscroll when within 100px of the bottom. This PR changes the threshold to 1 px to create a smooth transition in and out of autoscroll when a response is streaming.

Before:
https://github.com/user-attachments/assets/312eb168-0250-4b8a-80ea-0f81e439849b

After:
https://github.com/user-attachments/assets/a0e41022-c330-4b35-b36a-7fb852cec016

## How was this tested?

This was tested by changing the threshold value and prompting a long response to test the autoscroll while streaming, and scrolling up at different speeds. Low values above 1 px still resulted in noticeable jitter when scrolling slowly albeit much less than 100px, while setting it to 0 px made it nearly impossible to trigger autoscroll.

## Checklist

- [x] Frontend tests pass (npm test)
- [x] TypeScript passes (npx tsc --noEmit)
- [x] Frontend build passes (npm run build)

### DIFF
--- a/frontend/src/components/Chat/ChatArea.tsx
+++ b/frontend/src/components/Chat/ChatArea.tsx
@@ -22,6 +22,8 @@ export function ChatArea() {
   const navigate = useNavigate();
   const listRef = useRef<HTMLDivElement>(null);
   const shouldAutoScroll = useRef(true);
+  const wasStreaming = useRef(false);
+  const lastScrollTop = useRef(0);
 
   // Check if any data sources are connected
   const [hasConnectedSources, setHasConnectedSources] = useState<boolean | null>(null);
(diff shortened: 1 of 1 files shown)


ElliotSlusky (approved):
Verified the jitter mechanics and pushed a follow-up commit: disengage is now direction-based (any upward scroll stops autoscroll instantly), re-engage tolerates sub-pixel zoom residue (< 2px), and sending a message pins the view to the bottom even if you'd scrolled up. tsc, vitest (6/6), and the production build pass locally.

Issue #645, opened 2026-07-17
Unit tests failing


cesschneider:
### Description

When running the command `uv run pytest tests/ -v` I am getting a ModuleNotFoundError for modules polars and fastapi.

### Steps to Reproduce

git clone https://github.com/open-jarvis/OpenJarvis.git
cd OpenJarvis
uv sync --extra dev
uv run pre-commit install
uv run pytest tests/ -v

### Expected Behavior

All tests passed, and no errors.

### Actual Behavior

ERROR tests/evals/comparison/test_export_to_table_gen_roundtrip.py
ERROR tests/server/test_research_planner.py

### Operating System

macOS

### Python Version

3.11

### Hardware

Apple Silicon

### Engine

None

### Logs / Traceback

```shell
======================================================================================== ERRORS =========================================================================================
_____________________________________________________ ERROR collecting tests/evals/comparison/test_export_to_table_gen_roundtrip.py _____________________________________________________
ImportError while importing test module '/Users/cesar.schneider@efinancial.com/Github/cesschneider/openjarvis/tests/evals/comparison/test_export_to_table_gen_roundtrip.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
src/openjarvis/evals/comparison/table_gen.py:21: in <module>
    import polars as pl
E   ModuleNotFoundError: No module named 'polars'

The above exception was the direct cause of the following exception:
../../../.local/share/uv/python/cpython-3.11.13-macos-aarch64-none/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
tests/evals/comparison/test_export_to_table_gen_roundtrip.py:17: in <module>
    from openjarvis.evals.comparison.table_gen import (
src/openjarvis/evals/comparison/table_gen.py:23: in <module>
    raise ImportError(
E   ImportError: polars is required for the framework-comparison table generator. Install with: uv sync --extra framework-comparison
________________________________________________________________ ERROR collecting tests/server/test_research_planner.py _________________________________________________________________
ImportError while importing test module '/Users/cesar.schneider@efinancial.com/Github/cesschneider/openjarvis/tests/server/test_research_planner.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
[...]

PR #647, opened 2026-07-17
fix: make pytest suite hermetic against local dev-machine state

cesschneider:
## Summary

Running `uv run pytest tests/ -v` on a fresh dev machine (after building the mandatory `openjarvis_rust` extension) left 13 failures/6 errors that don't reproduce in CI, because CI's environment happens to sidestep two hermeticity gaps in the test suite:

- **CLI update-check nag pollutes captured output.** `check_for_updates()` writes its "new version available" banner to stderr, which Click's `CliRunner` merges into `result.output`. It already no-ops under `CI=true`, but a plain local run (network access + a version-check cache showing a newer release) gets the banner appended after JSON/CSV output, breaking `json.loads(result.output)` in `doctor --json`, `config show --json`, `ask --json`, `telemetry export --json/--csv`, etc.
- **`test_dense.py`'s Ollama skip-guard only checks TCP connectivity**, not whether the `nomic-embed-text` embed model is actually pulled. On a machine that runs Ollama for chat models but hasn't pulled that specific embed model, the retrieval-quality tests error out instead of skipping as the module docstring promises ("skipped if the server is unreachable").

## Changes

- `tests/conftest.py`: new autouse fixture sets `OPENJARVIS_NO_UPDATE_CHECK=1` for the whole suite, so the update-check nag never fires locally either.
- `tests/tools/storage/test_dense.py`: `_ollama_up()` now hits `/api/tags` and checks that `nomic-embed-text` is present, instead of a bare `socket.create_connection` port check.

No production code changed — this is purely test-isolation. Confirmed green with the suite's canonical CI invocation:

```
uv run pytest tests/ -n auto -q --tb=short -m "not live and not cloud and not hub"
# 7201 passed, 56 skipped, 0 failed
```
## Related issue
[Unit tests failing](https://github.com/open-jarvis/OpenJarvis/issues/645)

## Test plan

- [x] `uv run pytest tests/ -n auto -q -m "not live and not cloud and not hub"` — 7201 passed, 56 skipped
- [x] `uv run ruff check tests/conftest.py tests/tools/storage/test_dense.py` — clean
- [x] `uv run ruff format --check tests/conftest.py tests/tools/storage/test_dense.py` — clean
- [x] Verified `tests/cli/test_version_check.py` (which manages `OPENJARVIS_NO_UPDATE_CHECK` itself) still passes unaffected, since its local `_clean_env` fixture overrides the new global default.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -31,6 +31,19 @@ from openjarvis.core.registry import (
 
 
 @pytest.fixture(autouse=True)
+def _no_update_check(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Never let the CLI's PyPI update-check nag run during tests.
+
+    ``check_for_updates`` writes its banner to stderr, which ``CliRunner``
+    merges into ``result.output`` — polluting JSON/CSV output of any test
+    that invokes a CLI command. It already self-disables when ``CI`` is
+    set, but that only helps in CI; locally (e.g. a dev with a stale
+    version-check cache and network access) it fires for real.
+    """
+    monkeypatch.setenv("OPENJARVIS_NO_UPDATE_CHECK", "1")
+
+
+@pytest.fixture(autouse=True)
 def _clean_registries() -> None:
     """Ensure each test starts with empty registries and a fresh event bus."""
     ModelRegistry.clear()
--- a/tests/tools/storage/test_dense.py
+++ b/tests/tools/storage/test_dense.py
@@ -25,11 +25,28 @@ from openjarvis.tools.storage.dense import (
     chunk_markdown,
     dedupe_chunks,
 )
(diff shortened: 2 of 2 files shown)


ElliotSlusky (approved):
Both fixes verified: the update-check suppression is correct (the dedicated version-check tests manage the env var themselves and all 33 pass), and the model-aware Ollama probe reproduces correctly on a machine running Ollama without nomic-embed-text (13 passed / 9 skipped instead of errors). Pushed a follow-up: the probe now normalizes OLLAMA_HOST's documented full-URL form (it previously built http://<url>:<port>/api/tags), and the Ollama-backed tests construct DenseMemory against the same probed endpoint instead of the embedder's hard-coded localhost default, with unit tests for the probe itself.

PR #648, opened 2026-07-17
fix(desktop): preserve active chat when switching models

goatoush:
## Problem

Switching models from the desktop command palette calls `createConversation()` whenever the selected model changes which creates an empty conversation and a "New chat" entry each time. This can clutter your sidebar when switching models without sending messages as it gets full of "New chat" entries, and since it is possible to switch models within a chat, creating a new conversation each time just creates friction.

## Fix

Remove the conversation-creation side effect from model selection. Switching models now preserves the active chat while retaining the existing model preloading, loading state, logging, and in-flight stream cancellation behavior.

The next request uses the newly selected model with the current conversation context.

## Verification

- `npm run build` — passed
- `npm test` — 6 tests passed
- `npx tsc --noEmit` — passed
- `git diff --check` — passed

No public API or documentation changes are required.

### DIFF
--- a/frontend/src/components/CommandPalette.tsx
+++ b/frontend/src/components/CommandPalette.tsx
@@ -149,8 +149,7 @@ export function CommandPalette() {
     setCommandPaletteOpen(false);
 
     if (modelId !== previousModel) {
-      const { createConversation, setModelLoading, addLogEntry } = useAppStore.getState();
-      createConversation(modelId);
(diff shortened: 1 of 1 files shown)


ElliotSlusky (approved):
Verified locally: tsc, vitest (6/6), and production build all pass on the branch. Traced all model-selection paths — this also fixes a mid-stream state corruption where switching models could clobber the new chat's view with the old conversation's messages.

PR #655, opened 2026-07-21
fix(frontend): wrap tool descriptions in agent creator

goatoush:
## Problem

Tool descriptions would truncate instead of wrapping when hovering over a tool in the agent creation window, preventing you from seeing what many tools do.

## Summary

- allow tool descriptions in the agent creation window to wrap onto multiple lines
- align the description row correctly when text spans multiple lines

## Testing

- `npm test`
- `npx tsc --noEmit`
- `npm run build`


ElliotSlusky:
Thanks!

PR #656, opened 2026-07-21
fix(security): harden credentials, injection scanner, sandbox, and shell_exec

coutinho324-IA:
## What this PR does

Four small, self-contained defence-in-depth robustness fixes to the security layer, each with regression tests. No new dependencies.

### 1. `core/credentials.py` — TOML escaping
Credential values were serialized with naive f-string formatting (`f'{k} = "{v}"'`), which corrupts `credentials.toml` if a value contains a quote, backslash, or newline (e.g. `EMAIL_PASSWORD`). Now serialized via `tomlkit` (already a dependency), so values round-trip intact.

### 2. `security/injection_scanner.py` — Python fallback
`InjectionScanner` always called the Rust backend and crashed with `ImportError` when the compiled extension was unavailable (e.g. during the background build window on a fresh install). It also compiled a set of Python patterns that were never used. Now mirrors the graceful fallback in `security/ssrf.py`: uses Rust when available, otherwise runs the (previously dead) Python patterns.

### 3. `security/subprocess_sandbox.py` — Windows portability
`run_sandboxed` / `kill_process_tree` used POSIX-only `os.setsid` / `os.killpg` / `SIGKILL`, which raise `AttributeError` on Windows — a platform the project officially supports (`install.ps1`, classifiers). Now branches per platform: `CREATE_NEW_PROCESS_GROUP` + `taskkill /F /T` on Windows, `setsid` / `killpg` on POSIX.

### 4. `tools/shell_exec.py` — catastrophic-command denylist
Added a narrow `_DANGEROUS_PATTERNS` denylist (same style as `_BLOCKED_PATTERNS` in `code_interpreter.py` / `repl.py`), checked **before both execution paths** so the Rust backend cannot bypass it. Blocks only irreversible operations: root/home/wildcard `rm -rf`, fork bombs, raw block-device writes, `curl|sh`, recursive `chmod`/`chown` of `/`. Deliberately narrow — `rm -rf build/`, `git clean -fdx`, etc. still run.

## Testing

- Regression tests added/extended in `tests/core/test_credentials.py`, `tests/security/test_injection_scanner.py`, `tests/security/test_subprocess_sandbox.py`, `tests/tools/test_shell_exec.py`.
- The shell_exec denylist was validated against 24 commands (14 catastrophic + 10 legitimate) with zero false-positives.
- Added `scripts/validate_security_fixes.py` — exercises all four fixes without building the Rust extension or installing the full dependency set (all 8 checks pass on Windows).

## Notes for reviewers

Per `REVIEW.md`: no new components/registries introduced; changes are localized to existing security-boundary code; no hardcoded secrets. The denylist is defence-in-depth on top of the existing `code:execute` capability gate and `requires_confirmation`, not a sandbox.


ElliotSlusky:
Maintainer repair pushed in 3babd5a60cf6db48b78d1ae2c3389b73bb2d9870 after append-only integration of current main.

The original four fixes are retained and completed:
- credentials now serialize with tomlkit and use a same-directory atomic 0600 replacement, so escaping is correct and interrupted writes cannot truncate the existing store
- injection scanning falls back when Rust is absent, stale, fails to construct, or fails during a scan; a broken instance is disabled after the first failure so protection remains active without repeated tracebacks
- subprocess cleanup uses Windows process groups plus taskkill /T, POSIX start_new_session instead of unsafe preexec_fn, full process-group termination, bounded reap fallback, and the Windows environment variables required by cmd.exe
- shell_exec now always goes through the existing sandbox runner. This closes the prior Rust-path gaps where timeout, environment isolation, output limits, and nonzero exit status were not actually enforced. Common catastrophic direct commands for Unix and Windows are blocked before execution; capability checks and user confirmation remain the authorization boundary.
- malformed command, working-directory, and env_passthrough inputs now fail cleanly
- the standalone validation script was removed in favor of repository-native regression tests

Verification on the integrated current-main tree:
- 67 passed in the focused credential/injection/sandbox/shell suites
- 547 passed across all security tests plus credential, shell, code-interpreter, and REPL suites
- Ruff 0.12.12 check passed on every changed Python file
- Ruff format check passed on every changed Python file

The branch state was otherwise left unchanged.

ElliotSlusky (approved):

ElliotSlusky:
Closing with an invitation to split and resubmit. The underlying security goals are genuine, but this branch combines four independent changes, overlaps the now-merged memory injection-scanner work, predates the merged Windows environment/process fixes, and its destructive-command matcher still misses quoted HOME wipes. The credentials hardening and sandbox/shell hardening should be separate current-main PRs with adversarial tests. The source branch has not been deleted.

PR #657, opened 2026-07-21
fix(cli): make daemon/mine process management cross-platform (Windows)

coutinho324-IA:
## What this PR does

`jarvis daemon stop|restart` and `jarvis mine` use POSIX-only process control that breaks on the officially-supported Windows target (`install.ps1`, `Programming Language :: Python` classifiers, native-Windows install docs).

Two concrete failures on Windows:

1. **`signal.SIGKILL` does not exist on Windows** → `AttributeError` on the force-kill path in `daemon_cmd.stop()` and `mine_cmd._terminate_pid()`.
2. **`os.kill(pid, 0)` is not a safe liveness probe on Windows.** There `os.kill` maps *any* signal to `TerminateProcess`, so the "is it still running?" checks in `_read_pid()` / `_pid_alive()` would actually **kill** the process they were only meant to inspect.

## Approach

Adds two dependency-free helpers to `openjarvis/core/utils.py` (the module already hosts the cross-platform `get_python_executable` / `open_browser` helpers):

- **`process_alive(pid)`** — non-destructive liveness check. `tasklist` query on Windows, `os.kill(pid, 0)` on POSIX.
- **`terminate_process(pid, grace_seconds)`** — graceful then forced termination. `SIGTERM`→`SIGKILL` on POSIX; `taskkill`→`taskkill /F /T` on Windows. `signal.SIGKILL` is only referenced inside the POSIX branch, so it is never evaluated on Windows.

`daemon_cmd.py` and `mine_cmd.py` now delegate to these; the behaviour on POSIX is unchanged.

## Testing

- New `tests/core/test_utils.py`: liveness of the current/bogus/None/negative pid, and graceful+forced termination of a real spawned child on the host platform.
- Verified live on Windows: spawns a real child, confirms `process_alive` True, `terminate_process` kills it, `process_alive` then False; bogus/None pids never raise.

## Notes for reviewers

Independent of #656 (security hardening) — different concern (CLI process management portability), branched from `main`. No new dependencies. The SWE-bench eval harnesses (`mini_swe_agent.py`, `swebench_harness.py`) also use `os.killpg`/`SIGKILL` but are Linux/Docker-centric research tooling and are intentionally left out of scope.


ElliotSlusky:
Maintainer repair pushed at `564d9ae` after reconciling the branch with current `main`.

What changed:
- centralized daemon and mining PID handling in the shared cross-platform utilities;
- retained the signal-free Windows `OpenProcess`/`WaitForSingleObject` liveness probe already established on `main`;
- made POSIX liveness treat unreaped zombies as exited (the old `kill(pid, 0)` loop could wait and escalate forever);
- kept graceful-then-forced termination semantics (`SIGTERM`/`SIGKILL` on POSIX, `taskkill` then `/F /T` on Windows);
- added simulated Windows coverage for missing, access-denied, exited, and live handles, plus graceful and forced `taskkill` command paths.

Verification:
- Ruff check and format check: clean.
- Focused process/daemon/mining tests: **49 passed**.
- Full `tests/core` + `tests/cli`: **736 passed, 2 skipped**. The two remaining memory-index failures reproduce unchanged on current `main` with this environment's stale Rust extension (`SQLiteMemory.replace_source` is absent), so they are not caused by this PR.
- The update passed an exact open-state/head-SHA/maintainer-modifiable gate and was pushed non-force to the contributor branch.

ElliotSlusky (dismissed):
Re-reviewed after the current-main reconciliation and process-lifecycle hardening. The focused regression suite passes, Windows-only paths now have deterministic unit coverage, and the broad-suite delta is clean.

PR #658, opened 2026-07-21
feat(analytics): honor DO_NOT_TRACK env kill-switch for opt-out

coutinho324-IA:
## What this PR does

Analytics is on by default (`AnalyticsConfig.enabled = True`). For a project whose thesis is *"Personal AI, On Personal Devices"*, respecting the console [`DO_NOT_TRACK` standard](https://consoledonottrack.com) gives users an **environment-level opt-out** that doesn't require editing config files — the convention many CLI tools already honor, and handy for CI, shared machines, or a global shell default.

## Changes

- **`analytics/identity.py`**: `is_analytics_enabled()` now returns `False` when `DO_NOT_TRACK` is set to `1` / `true` / `yes` (case-insensitive), regardless of config. `0` / `false` / unset defer to config exactly as before, so existing behaviour is unchanged unless the user explicitly sets the kill-switch. Adds a small public `do_not_track()` helper.
- **`docs/telemetry.md`**: the intro promised "…and how to opt out" but the page had no opt-out section. Adds a **How to opt out** section documenting both `DO_NOT_TRACK` and the existing `[analytics] enabled = false` config flag.
- **`tests/analytics/test_identity.py`** (new): covers the kill-switch (truthy/falsey/unset values), config precedence (DNT overrides an enabled config; `DNT=0` does not disable), and anon-id persistence/reset.

## Testing

Validated the full truth table in isolation (19 checks): DNT truthy values disable analytics even when config is enabled; `0`/`false`/`""`/unset defer to config; anon-id round-trips and resets.

## Notes for reviewers

Independent of #656 / #657. No new dependencies, no data-collection surface added — this only *narrows* what is collected. The change is fail-safe: any ambiguity resolves toward *not* tracking.

### DIFF
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -14,6 +14,19 @@ collected, where the data goes, and how to opt out.
   data is not sold or shared with third parties.
 - **365-day retention**, after which events are deleted automatically.
 
+## How to opt out
+
+Analytics is on by default. Either of these turns it off — when either is
+active, no analytics client is initialised and no events are sent:
+
+- **Environment kill-switch (no config edit):** set `DO_NOT_TRACK=1` (or
+  `true`/`yes`). This honors the console
+  [`DO_NOT_TRACK` standard](https://consoledonottrack.com) and disables
+  analytics regardless of config — handy for CI, shared machines, or a global
+  shell default.
+- **Config:** set `enabled = false` under `[analytics]` in
+  `~/.openjarvis/config.toml`.
+
 ## What we collect
 
 ### Lifecycle events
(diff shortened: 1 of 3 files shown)


ElliotSlusky:
Thank you for your submission. After further consideration, we have decided not to implement support for the DO_NOT_TRACK specification as part of this pull request. we're going to decline merging your contribution and moving forward with analytics configuration support within OpenJarvis rather than creating an additional environment-level feature. We appreciate your effort in submitting and testing your proposal. Thank you for your interest in our project.

PR #683, opened 2026-07-27
fix: bind a deep_research agent's configured tools on the server path

### LINKED ISSUE #682: deep_research agents silently ignore their configured tools on the server path
### Description

A managed agent of type `deep_research` is bound only to the four hard-coded knowledge-base tools built by `_build_deep_research_tools()`. Its own `config["tools"]` — whatever the user selected in the agent-creation wizard — is never read, and MCP tools are never merged either.

The cause is control flow in `src/openjarvis/server/agent_manager_routes.py`. `_stream_managed_agent()` branches on `agent_type == "deep_research"` at [L865](https://github.com/open-jarvis/OpenJarvis/blob/main/src/openjarvis/server/agent_manager_routes.py#L865), builds `dr_tools`, and at [L889](https://github.com/open-jarvis/OpenJarvis/blob/main/src/openjarvis/server/agent_manager_routes.py#L889) returns a `StreamingResponse` that runs `DeepResearchAgent(tools=dr_tools, ...)`. The generic tool resolution lives *below* that early return — `_resolve_tool_specs(config.get("tools"))` at L1169 and `_get_mcp_tools()` at L1181 — so neither is ever reached for this agent type.

Nothing warns. The agent is created successfully, the UI shows the selected tools, and the tools simply do not exist at run time. The create-agent wizard writes `config.tools` for every agent type (`frontend/src/pages/AgentsPage.tsx#L742`) and `CreateAgentRequest.config` is free-form, so this is silently-ignored configuration rather than an unsupported option.

The CLI path is unaffected, which is probably why this hasn't surfaced.

Related detail: `_build_deep_research_tools()` returns `[]` when `~/.openjarvis/knowledge.db` does not exist, and the `if dr_tools:` guard at L890 then falls through to the generic path — so a `deep_research` agent *does* get its configured tools, but only on machines with no knowledge DB. Same config, opposite behaviour, depending on whether a file exists.

### Steps to Reproduce

1. Start the server: `uv run jarvis serve`
2. In the web UI, create an agent with **Agent Type = `deep_research`** and, in the tools step, tick `web_search` (or any non-knowledge tool) in addition to the defaults. Equivalently: `POST /api/agents` with `{"agent_type": "deep_research", "config": {"tools": ["knowledge_search", "web_search", "think"]}}`.
3. Send it a message that requires the web, e.g. *"What did OpenAI announce this week? Search the web."*
4. Inspect what the agent was actually able to call:

```sql
-- ~/.openjarvis/agents.db
SELECT tool_calls FROM agent_messages WHERE agent_id = '<id>';
SELECT event_type, content FROM agent_learning_log WHERE agent_id = '<id>';
```

### Expected Behavior

The agent is bound to the union of its configured tools, the knowledge-base tools, and any discovered MCP tools — the same set the create-agent wizard showed the user. If a configured tool cannot be resolved, a warning is logged, matching `_resolve_tool_specs()`'s existing behaviour for unknown names.

### Actual Behavior

The agent is bound to exactly four tools (`knowledge_search`, `knowledge_sql`, `scan_chunks`, `think`). `web_search` and every other configured or MCP tool is absent, with no warning in the server log and no indication in the UI.

Because the model is asked to search and has nothing to search with, it answers from parametric memory instead. On my install, every one of the 47 recorded messages for such an agent had `tool_calls = NULL`, and the answers contained plausible-looking but unverifiable references.

### Operating System

Windows (the code path is platform-independent; not Windows-specific)

### Python Version

3.12

### Hardware

NVIDIA GPU

### Engine

Ollama

### Logs / Traceback

```shell
No traceback — the failure is silent. The server log shows the agent
starting and streaming normally; there is no "unknown tool" warning
because config["tools"] is never inspected on this path.
```

---

### Proposed fix

Resolve the agent's configured tool *names* into live tool *instances* before constructing `DeepResearchAgent`, and merge them with the knowledge tools and the MCP adapters rather than replacing them.

That needs a companion to `_resolve_tool_specs()` — call it `_resolve_tool_instances()` — that reads names identically (browser meta-tool expansion, skipping channel tools, warning on unknown names) but returns instantiated tools, which is what an agent running its own loop needs, as opposed to the spec dicts handed to the engine.

I have this implemented and tested against a live agent: the bound toolset goes from 4 to 15, and `agent_learning_log` records a real `web_search` call and result where it previously recorded none. Happy to open a PR if the approach looks right — raising it here first since it touches the managed-agent tool wiring, per CONTRIBUTING.

One deliberate non-change: conversation history is still not replayed on this path. That's a separate concern and I'd rather not fold it in.

Curryraj:
Fixes #682.

Through the server path a managed `deep_research` agent received only the four knowledge-base tools from `_build_deep_research_tools()`. Its own `config["tools"]` was never read, because `_resolve_tool_specs()` sits further down the generic path, past the early return in the `deep_research` branch; MCP tools were dropped for the same reason.

The agent-creation wizard writes `config.tools` for every agent type and `CreateAgentRequest.config` is free-form, so this is silently discarded configuration rather than an unsupported option: the agent is created, the UI shows the selected tools, and they do not exist at run time. Nothing is logged, because the config is never inspected on this path. The CLI path is unaffected, which is why it went unnoticed.

## Approach

Adds `_resolve_tool_instances()`, which mirrors `_resolve_tool_specs()` in how it reads names — `browser` meta-tool expansion, skipping channels, warning on unknown names — but returns live instances rather than the OpenAI-format spec dicts handed to the engine. An agent running its own loop needs instances.

Those are merged with the knowledge tools and the MCP adapters rather than replacing them, so local retrieval and web search are both available.

## Verification

Against a live agent, the bound toolset goes from 4 tools to 15, and `agent_learning_log` records a real `web_search` call and result where it previously recorded none.

`tests/server/test_deep_research_tools_wiring.py` — 6 passed. Full `tests/server` — 294 passed, 4 failed; all 4 failures reproduce identically on `main` without this change and are addressed in two separate PRs.

## Note

Conversation history is deliberately still not replayed on this path. The comment at the call site records the measurement behind that — on `qwen2.5:7b`, replaying the stored history made the agent stop calling tools entirely. Re-introducing memory there needs history that models the desired behaviour, and belongs in its own change.

### DIFF
--- a/src/openjarvis/server/agent_manager_routes.py
+++ b/src/openjarvis/server/agent_manager_routes.py
@@ -588,6 +588,78 @@ def _instantiate_managed_tool(
     return tool_cls()
 
 
+def _tool_instance_name(tool: Any) -> str:
+    """Best-effort name for a tool instance (empty string if unavailable)."""
+    try:
+        return tool.spec.name
+    except Exception:
+        return ""
+
+
+def _resolve_tool_instances(
+    tool_config: Any,
+    *,
+    engine: Any,
+    model: str,
+    app_state: Any,
+) -> List[Any]:
+    """Instantiate a managed agent's configured tools.
+
+    Mirrors :func:`_resolve_tool_specs` in how it interprets names (``browser``
+    meta-tool expansion, silently skipping channels, warning on unknown names)
+    but returns live tool *instances* rather than OpenAI-format specs.
+
+    Agent classes that run their own loop — ``DeepResearchAgent`` in particular
+    — take real instances, whereas the raw streaming path hands specs to the
+    engine. Without this, an agent's configured tools cannot reach it at all.
+    """
+    if not tool_config:
+        return []
+
+    from openjarvis.core.registry import ChannelRegistry, ToolRegistry
+
+    _ensure_registries_populated()
+
+    instances: List[Any] = []
+    seen: set = set()
+
+    def _add(name: str) -> None:
+        if name in seen or not ToolRegistry.contains(name):
+            return
+        try:
+            instances.append(
+                _instantiate_managed_tool(
+                    ToolRegistry.get(name),
+                    name,
+                    engine=engine,
+                    model=model,
+                    app_state=app_state,
+                )
+            )
+            seen.add(name)
+        except Exception as exc:
+            logger.warning("Could not instantiate tool '%s' (%s) — dropping", name, exc)
+
+    for entry in tool_config:
+        if not isinstance(entry, str):
+            continue
+        if entry == "browser":
+            for sub in _BROWSER_SUB_TOOLS:
+                _add(sub)
+            continue
+        if ChannelRegistry.contains(entry):
+            continue
+        if not ToolRegistry.contains(entry):
+            logger.warning(
+                "Tool '%s' referenced in agent config but not in ToolRegistry",
+                entry,
+            )
+            continue
+        _add(entry)
+
+    return instances
+
+
 def _build_deep_research_tools(
     engine: Any,
     model: str,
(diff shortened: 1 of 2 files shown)


ElliotSlusky:
Thank you @Curryraj. I took lots of what you did and fixed the issue.

PR #715, opened 2026-08-10
fix(security): block sensitive file aliases

JustInagap:
## Summary

- evaluate both lexical paths and resolved targets in the sensitive-file policy
- fail closed when a target cannot be resolved safely while permitting missing write leaves
- make Python and Rust matching case-insensitive and restore `*.env` backend parity
- cover Python read, write, and patch tools plus Rust-native read/write consumers

## Root cause

The policy checked only the submitted filename. A safe-looking symlink such as `notes.txt` could therefore point at `.env`, and case variants behaved differently across platforms. The Rust matcher also omitted the Python catalog's `*.env` rule.

## Impact

Stable symlink aliases and case variants can no longer bypass sensitive-file restrictions through either the Python or Rust-native tool paths. Public APIs and dependency sets remain unchanged.

## Validation

- 121 focused Python tests passed
- standalone changed Rust policy compiled under `-D warnings`; 3 tests passed
- Ruff and rustfmt checks passed
- Unix-only Rust integration tests cover native read/write denial and target preservation
- independent re-review found no Critical, Important, or Minor issues

Full Cargo workspace checking was unavailable on the Windows development host because it lacks a functioning native linker; the Unix-only native integration test is expected to run in CI.

### DIFF
--- a/rust/crates/openjarvis-security/src/file_policy.rs
+++ b/rust/crates/openjarvis-security/src/file_policy.rs
@@ -1,39 +1,31 @@
 //! File sensitivity policy — block access to secrets, credentials, and keys.
 
-use once_cell::sync::Lazy;
-use std::collections::HashSet;
-use std::path::Path;
+use std::ffi::OsString;
+use std::io;
+use std::path::{Path, PathBuf};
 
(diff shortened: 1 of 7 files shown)


ElliotSlusky:
Hey, thank you for the work! Please resubmit when you're ready for review.

PR #716, opened 2026-08-10
test: make dev test collection hermetic

JustInagap:
## Summary

- skip comparison and server test modules only when their declared optional dependencies are absent
- replace live DuckDuckGo fallback calls with a deterministic fixture at the network boundary
- keep production imports and real `ToolResult` construction active whenever dependencies are installed

## Why

A normal development-only environment could not collect the full test suite because two test modules imported optional packages unconditionally. Several fallback tests could also escape to the public DuckDuckGo service, making local and CI results depend on network availability.

## Impact

Contributors can collect the suite with the documented development dependencies, while optional-extra CI lanes still execute the guarded modules normally. Web-search fallback tests are deterministic and offline without over-mocking the tool's control flow.

## Validation

- `pytest tests/tools/test_web_search.py`: 38 passed
- full development-only collection: 7,079 tests collected
- optional dependencies installed: 57 focused tests passed
- Ruff check and format check passed
- independent review found no Critical, Important, or Minor issues

### DIFF
--- a/tests/evals/comparison/test_export_to_table_gen_roundtrip.py
+++ b/tests/evals/comparison/test_export_to_table_gen_roundtrip.py
@@ -14,6 +14,13 @@ from __future__ import annotations
 import json
 from pathlib import Path
 
+import pytest
+
+pytest.importorskip(
+    "polars",
+    reason="framework-comparison extra not installed",
+)
+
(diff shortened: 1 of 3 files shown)


ElliotSlusky:
Hey! Please resubmit when you're ready for review.

PR #717, opened 2026-08-10
fix(cli): preserve environment during self-update

### LINKED ISSUE #712: self-update silently uninstalls desktop/desktop-native extras (fastapi, openjarvis-rust, faster-whisper, ...)
## Description

`jarvis self-update` (and its underlying suggested command, `git pull && uv sync`) silently **uninstalls** packages belonging to extras/dependency-groups that were used at install time but are not implied by a bare `uv sync`. For an editable-git desktop install (`uv sync --extra desktop --group desktop-native`), this strips out `openjarvis-rust`, `fastapi`, `starlette`, `uvicorn`, `faster-whisper`, `av`, `ctranslate2`, `onnxruntime`, and more — breaking `jarvis ask` / the REST API / speech backend — with no warning and no automatic reinstall.

## Repro (Windows 11, native install via install.ps1)

```
PS> jarvis self-update
Current OpenJarvis version: v0.0.1.dev1+unknown.g95a985798
Install method: editable-git
Upgrade command: cd C:\Users\<user>\AppData\Local\OpenJarvis\src && git pull && uv sync

Run the upgrade command now? [Y/n]: y

→ cd ... && git pull && uv sync

Already up to date.
Resolved 386 packages in 3ms
Uninstalled 15 packages in 2.65s
 - av==16.1.0
 - ctranslate2==4.7.1
 - fastapi==0.129.0
 - faster-whisper==1.2.1
 - flatbuffers==25.12.19
 - mpmath==1.3.0
 - onnxruntime==1.24.2
 - openjarvis-rust==0.1.0 (from file:///.../rust/crates/openjarvis-python)
 - protobuf==6.33.5
 - python-multipart==0.0.22
 - setuptools==82.0.0
 - starlette==0.52.1
 - sympy==1.14.0
 - tokenizers==0.22.2
 - uvicorn==0.41.0

Upgrade complete. Re-run `jarvis --version` to confirm.
```

After this, `jarvis ask` fails (the native extension / FastAPI server / speech backend are gone). Only visible fix is to manually re-run:

```
uv sync --extra desktop --group desktop-native
```

## Root cause

`jarvis self-update` (and `install.ps1` / `install.sh` docs) always suggest plain `git pull && uv sync`, which doesn't know which extras/groups the environment was originally synced with. `uv sync` without `--extra`/`--group` flags resolves/installs only the base dependency set, and — since it also reconciles the venv to match that smaller set — it actively **removes** anything from a previous sync that isn't in the base set.

This is the same log pattern visible (as an aside) in #478, though that issue focused on the version-string-not-changing symptom rather than the silent extras removal.

## Suggested fix

`jarvis self-update` should remember (e.g. persist to the venv metadata or a marker file) which `--extra`/`--group` flags were used at install time, and re-apply them on every `uv sync` it runs — or at minimum warn loudly before uninstalling desktop-native packages so users aren't left with a silently broken `jarvis ask`.

## Environment

- Windows 11 (build 26200), native `install.ps1` install (not WSL)
- Python 3.11
- uv, editable-git install (`openjarvis==0.0.1.dev1+unknown.g95a985798`, HEAD of `main`)
- Extras used at install: `--extra desktop --group desktop-native`

JustInagap:
Closes #712.

## Summary

- persist the extras and dependency groups selected by official editable installers
- reproduce that profile during project-venv self-updates and preserve packages when no valid profile exists
- reinstall external editable environments into their active Python interpreter
- execute packaged updates from trusted argv instead of parsing display text
- stop before dependency changes when `git pull --ff-only` fails

## Root cause

Editable installs were treated as one generic case. Self-update ran an exact `uv sync` without knowing which optional features the installer selected, so valid packages could be removed. The command also did not distinguish the repository's own `.venv` from an external editable environment.

## Impact

Official Windows and quickstart installations record a small validated profile that future updates reproduce. Existing checkouts without a profile use `uv sync --inexact` with a warning, avoiding destructive package pruning. PyPI and uv-tool display strings are now presentation-only and never become execution input.

## Validation

- 40 focused CLI, profile, and packaging tests passed
- Ruff check and format check passed
- Git Bash syntax validation passed for `scripts/quickstart.sh`
- PowerShell AST validation passed for `deploy/windows/install.ps1`
- independent re-review found no Critical, Important, or Minor issues

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -130,3 +130,6 @@ learning.db
 minion_logs/
 *.oj-debug.json
 oj-debug.*.json
+
+# Local editable-install feature profile (written by official installers)
+/.openjarvis-install-profile.json
--- a/deploy/windows/install.ps1
+++ b/deploy/windows/install.ps1
@@ -289,6 +289,11 @@ try {
     if ($LASTEXITCODE -ne 0) {
         Write-Fail "uv sync failed with exit code $LASTEXITCODE. Check the output above."
     }
+    & $uvExe run --no-sync python -m openjarvis.cli._install_profile record `
+        --extra desktop --group desktop-native
+    if ($LASTEXITCODE -ne 0) {
+        Write-Fail "Failed to record the editable install profile."
+    }
 } finally {
     Pop-Location
 }
--- a/docs/user-guide/cli.md
+++ b/docs/user-guide/cli.md
@@ -9,6 +9,27 @@ jarvis --version   # Print the OpenJarvis version
 jarvis --help      # Show top-level help with all subcommands
 ```
 
+## `jarvis self-update`
+
+Upgrade OpenJarvis using the active installation method:
+
+```bash
+jarvis self-update --check  # Preview the detected command
+jarvis self-update          # Confirm, then update
+jarvis self-update --yes    # Update without prompting
+```
+
+PyPI and `uv tool` installs use their normal upgrade commands. Editable Git
+installs pull the checkout with a fast-forward-only update, then update the
+environment that is currently running OpenJarvis.
+
+Official project-local installers record their uv extras and dependency groups
+in a local, ignored install-profile file. Self-update reuses that exact feature
+set. If an older project-local checkout has no valid profile, self-update uses
+`uv sync --inexact` and prints a warning so optional packages are preserved.
+Editable installs in an external virtual environment are refreshed with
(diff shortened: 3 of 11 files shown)


ElliotSlusky:
Hey @JustInagap. Your PR had a few issues so I went ahead and fixed them https://github.com/open-jarvis/OpenJarvis/pull/718.

Thank you for this!

PR #743, opened 2026-08-15
fix(frontend): move MessageBubble hooks above the early return

ar1vit0r:
## Summary
- `MessageBubble`'s three `useMemo` calls ran after the `isUser` early return, so they were skipped entirely for user messages — a rules-of-hooks violation (hook call count differs between renders depending on `message.role`).
- Pure reordering, no logic change: the three `useMemo` calls now run unconditionally before the early return.

## Test plan
- [x] `tsc -b` — clean
- [x] `vitest run` — all tests pass


ElliotSlusky:
taking a look

ElliotSlusky:
LGTM thanks!

PR #745, opened 2026-08-15
fix(agents): repair ClaudeCodeAgent's Node runner build and auth precedence

ar1vit0r:
## Summary
- The runner imported `query()`/`Tool` from `@anthropic-ai/claude-code`, but that package has never published an importable SDK — it's the `claude` CLI binary only. Swapped to the real package, `@anthropic-ai/claude-agent-sdk`, rewrote the runner against its actual `AsyncGenerator<SDKMessage>` API, and added the `tsconfig.json`/build script that were missing entirely (the runner shipped as bare TypeScript source with nothing to compile it).
- Fixed the coupled auth-precedence bug: `ClaudeCodeAgent` no longer defaults `api_key` from the ambient `ANTHROPIC_API_KEY` env var, and the runner strips that var from the subprocess env unless a key was explicitly passed, so a Claude Pro/Max login session is used instead of silently billing metered API credits.
- `_ensure_runner()` now detects a stale `node_modules` (missing a newly-added devDependency) via mtime instead of just presence.
- Included a follow-up test-isolation fix: `get_config_dir()` checks `$OPENJARVIS_HOME`/`$XDG_DATA_HOME` before `Path.home()`, so on a machine with either set, the `TestEnsureRunner` tests' patched `Path.home()` was silently ignored and `_ensure_runner()` wrote into the real runner dir, breaking the `subprocess.run` call-count assertions. Also fixed the reinstall-staleness test backdating `node_modules`'s mtime relative to "now" instead of the actual source `package.json`'s mtime, which could silently never trigger the staleness check it exists to catch.

## Test plan
- [x] `pytest tests/agents/test_claude_code.py -q` — 28/28 pass
- [x] `npm install && npm run build` in `claude_code_runner/` — clean `tsc` build
- [x] `ruff check` / `ruff format --check` — clean

### DIFF
--- a/src/openjarvis/agents/claude_code.py
+++ b/src/openjarvis/agents/claude_code.py
@@ -1,6 +1,6 @@
 """ClaudeCodeAgent -- wraps the Claude Agent SDK via Node.js subprocess bridge.
 
-Spawns a Node.js runner process that calls the ``@anthropic-ai/claude-code``
+Spawns a Node.js runner process that calls the ``@anthropic-ai/claude-agent-sdk``
 SDK, communicating via JSON over stdin/stdout with sentinel-delimited output.
 
 The engine parameter is accepted for interface conformance with BaseAgent but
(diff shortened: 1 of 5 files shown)


ElliotSlusky:
Thank you for the PR. 

The code has several regressions. The documented API-key authentication stops working, first-run build fails under NODE_ENV=production, and SDK failures become “Unknown error.”

Please fix these or I can step in and fix.

ElliotSlusky:
I'm happy to fix. Fixing.

ElliotSlusky:
Thanks for the thorough fix, but we’ve decided to remove the unused ClaudeCodeAgent integration rather than continue maintaining it.

ar1vit0r:
Hey, sure! I understand! thx for the try  :D

PR #790, opened 2026-08-17
fix(cli): accept a filesystem path for `jarvis init --config` (#768)

### LINKED ISSUE #768: jarvis init --config <path> crashes with AttributeError
Description: The `--config` option is declared `type=click.Path(exists=True)` without `path_type=Path` (src/openjarvis/cli/init_cmd.py:232-235), so Click passes the callback a plain `str`, while the function body treats it as a `Path` (`config.read_text()`, `config.write_text()`). Every use of `--config` raises an unhandled `AttributeError`.

```python
@click.option(
    "--config",
    type=click.Path(exists=True),
    help="Path to config file to use.",
)
...
toml_content = config.read_text()
```

Steps to Reproduce:
1. Create any file, e.g. `some-config.toml`.
2. Run `jarvis init --config some-config.toml`.
3. Observe an unhandled `AttributeError: 'str' object has no attribute 'read_text'`.

Expected Behavior: `--config` accepts an existing file path and reads it successfully.

Actual Behavior: The command crashes with an `AttributeError` on any use of `--config`.

Suggested fix: Add `path_type=Path` to the `click.Path(...)` declaration (`type=click.Path(exists=True, dir_okay=False, path_type=Path)`) so Click hands back a real `Path` object matching what the function body expects. A local patch doing this, plus a regression test, is available if useful.

ethanhawkes-gif:
## Summary

Fixes #768. `jarvis init --config <file>` always crashed with
`AttributeError: 'str' object has no attribute 'read_text'`, so the `--config`
option was completely unusable.

## Root cause

The option is declared:

```python
@click.option(
    "--config",
    type=click.Path(exists=True),
    ...
)
```

`click.Path` without `path_type=Path` hands the callback a **`str`**, but the
`init()` body annotates the parameter `config: Optional[Path]` and uses it as a
`pathlib.Path`:

```python
if config:
    toml_content = config.read_text()   # init_cmd.py:448
...
if config:
    config.write_text(toml_content)     # init_cmd.py:457
```

`str` has no `read_text()`, so every invocation of `--config` raised an
unhandled `AttributeError`.

## Fix

```diff
-    type=click.Path(exists=True),
+    type=click.Path(exists=True, dir_okay=False, path_type=Path),
```

- `path_type=Path` makes Click pass a real `Path`, matching what the function
  body already expects (`Path` is already imported).
- `dir_okay=False` rejects a directory with a clean Click usage error instead
  of letting it fall through to another crash. Both flags follow the
  `click.Path(..., path_type=Path)` convention already used elsewhere in the
  CLI (e.g. `mine_cmd.py`).

Behavior for a valid file argument is unchanged apart from the callback now
receiving a `Path` (which is what it always assumed). The `--config`-omitted
path (`config is None`) is unaffected.

## Tests

Adds `tests/cli/test_init_config.py`:

- `test_init_config_reads_provided_file` — `jarvis init --config <file>` now
  exits 0 and reads the file (this **fails on `main`** with the exact reported
  `AttributeError` and passes with the fix).
- `test_init_config_rejects_directory` — a directory argument is rejected with
  exit code 2 rather than crashing.

Verified locally:

- `tests/cli/` — 479 passed, 4 skipped.
- Full lane `pytest tests/ -m "not live and not cloud and not hub"` (with the
  Rust extension built) — 7574 passed; the only failures were pre-existing and
  environment-dependent (cloud-key / credential-absence tests that fail when API
  keys are present in the environment), identical on an unmodified checkout.
- `ruff check` and `ruff format --check` clean on the changed files.

### DIFF
--- a/src/openjarvis/cli/init_cmd.py
+++ b/src/openjarvis/cli/init_cmd.py
@@ -230,7 +230,7 @@ def _do_download(engine: str, model: str, spec, console: Console) -> None:
 )
 @click.option(
     "--config",
-    type=click.Path(exists=True),
+    type=click.Path(exists=True, dir_okay=False, path_type=Path),
     help="Path to config file to use.",
 )
 @click.option(
--- a/src/openjarvis/cli/init_cmd.py
+++ b/src/openjarvis/cli/init_cmd.py
@@ -445,7 +445,7 @@ def init(
             )
 
     if config:
-        toml_content = config.read_text()
+        toml_content = config.read_text(encoding="utf-8")
     else:
         if full_config:
             toml_content = generate_default_toml(hw, engine=engine, host=host)
@@ -453,10 +453,10 @@ def init(
             toml_content = generate_minimal_toml(hw, engine=engine, host=host)
 
     DEFAULT_CONFIG_DIR.mkdir(parents=True, exist_ok=True)
-    if config:
-        config.write_text(toml_content)
-    else:
-        DEFAULT_CONFIG_PATH.write_text(toml_content)
+    # ``--config`` selects an existing file to install; it is not an alternate
+    # output path.  Always activate the selected/generated configuration at
+    # the canonical location that subsequent ``jarvis`` commands load.
+    DEFAULT_CONFIG_PATH.write_text(toml_content, encoding="utf-8")
 
     console.print()
     console.print(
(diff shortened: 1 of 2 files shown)


ElliotSlusky:
Maintainer repair pushed in `20a97a38`: `jarvis init --config` now installs and activates the supplied file at `DEFAULT_CONFIG_PATH`, computes the digest from the installed copy, and leaves the source untouched. Added coverage for canonical installation, source preservation, digest mode, and directory rejection. Verified: `pytest tests/cli/test_init_config.py -q` (3 passed), plus Ruff check/format.

ElliotSlusky (approved):

PR #793, opened 2026-08-17
Fix silent TTS failure in morning digest

kvbuuren:
Fix silent TTS failure in morning digest

Cartesia's default "sonic" model was sunsetted, causing every TTS call
to fail with a 400. The failure was masked because digest_cmd checked
Path.exists() on the audio path, and an empty path stringifies to ".",
which always exists as a directory. Switch the default model to
"sonic-2" and check is_file() instead so a failed TTS run is reported
accurately.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


kvbuuren:
Sluiten voor nu - niet klaar om dit publiekelijk te delen op dit moment.

PR #794, opened 2026-08-17
fix(tools): return promptly on tool execution timeout

### LINKED ISSUE #749: Tool execution timeout does not return until the tool actually finishes
Description: `ToolExecutor.execute()` runs each tool inside `with concurrent.futures.ThreadPoolExecutor(max_workers=1) as pool:` and calls `future.result(timeout=timeout)` (src/openjarvis/tools/_stubs.py:264-266). When that call raises `TimeoutError`, control still has to leave the `with` block, and `ThreadPoolExecutor.__exit__` calls `shutdown(wait=True)`, which blocks until the worker thread finishes running the tool. So a "timed out" `ToolResult` is only handed back once the tool has actually completed - the configured `timeout_seconds` has no effect on when the caller gets control back. This is most severe on the server, where `agent_manager_routes.py` calls `stream_tool_executor.execute()` synchronously from inside an `async def generate()` generator (src/openjarvis/server/agent_manager_routes.py:1449, enclosing function at line 1313), so the block runs on the asyncio event loop and freezes every other request for as long as the wedged tool keeps running.

```python
with concurrent.futures.ThreadPoolExecutor(max_workers=1) as pool:
    future = pool.submit(tool.execute, **params)
    result = future.result(timeout=timeout)
except concurrent.futures.TimeoutError:
    ...
```

Steps to Reproduce:
1. Register a tool whose `ToolSpec` sets `timeout_seconds=1.0` and whose `execute()` sleeps for 8 seconds.
2. Call `ToolExecutor([SlowTool()]).execute(ToolCall(name="slow", arguments="{}"))` and time it.
3. Observe the call returns after ~8 seconds with content `"Tool 'slow' timed out after 1s."`, not after ~1 second.

Expected Behavior: `execute()` returns within about `timeout_seconds` regardless of how long the underlying tool keeps running.

Actual Behavior: `execute()` blocks until the tool call finishes, then returns the "timed out" result - the timeout is cosmetic.

Environment: OpenJarvis @ 1ca3ec6 (main), Python 3.13.1, Windows 11. Platform-independent - reproduces on any OS.

Suggested fix: Build the `ThreadPoolExecutor` outside a `with` block (not as a context manager) and shut it down in a `finally` with `pool.shutdown(wait=False, cancel_futures=True)`, so the timeout branch returns immediately in every case. Note `cancel_futures` cannot cancel an already-running future, so the orphaned worker keeps running in the background (joined by `concurrent.futures`' atexit hook at interpreter shutdown) - tools that can block indefinitely still need their own internal deadline. A local patch implementing this, plus a regression test, is available if useful.

havolline1:
## Summary
- Fixes #749: `ToolExecutor.execute()` used `ThreadPoolExecutor` as a context manager, so on timeout the `with` block's `__exit__` still called `shutdown(wait=True)` and blocked until the slow tool actually finished — defeating the configured timeout and freezing the event loop on servers.
- Moves the pool out of the `with` block and shuts it down with `wait=False, cancel_futures=True` in a `finally` clause, so `execute()` returns promptly at the timeout instead of waiting for the tool to complete.
- Note (also called out in the issue): this cannot interrupt a tool already mid-execution — Python has no thread-kill — so a runaway tool still consumes a worker thread until it finishes on its own. Real interruption would need an internal deadline/cancellation mechanism inside each tool; out of scope here.

## Test plan
- [x] Added `test_slow_tool_times_out_promptly` (`tests/tools/test_tool_timeout.py`), which asserts `execute()` returns in <3s for a tool with `timeout_seconds=1.0` that sleeps 8s. Confirmed it fails at ~8.0s against the old code and passes after the fix.
- [x] `uv run pytest tests/tools/` — 666 passed, 13 skipped
- [x] `uv run pytest tests/security/test_tool_confirmation.py tests/security/test_security_wiring.py tests/security/test_boundary_guard.py` — 30 passed

### DIFF
--- a/src/openjarvis/tools/_stubs.py
+++ b/src/openjarvis/tools/_stubs.py
@@ -260,10 +260,10 @@ class ToolExecutor:
         # Execute with timeout
         timeout = tool.spec.timeout_seconds or self._default_timeout
         t0 = time.time()
+        pool = concurrent.futures.ThreadPoolExecutor(max_workers=1)
         try:
-            with concurrent.futures.ThreadPoolExecutor(max_workers=1) as pool:
-                future = pool.submit(tool.execute, **params)
-                result = future.result(timeout=timeout)
+            future = pool.submit(tool.execute, **params)
+            result = future.result(timeout=timeout)
         except concurrent.futures.TimeoutError:
             if self._bus:
                 self._bus.publish(
@@ -281,6 +281,12 @@ class ToolExecutor:
                 content=f"Tool execution error: {exc}",
                 success=False,
             )
+        finally:
+            # wait=False so a timed-out call returns immediately instead of
+            # blocking on the still-running worker thread. cancel_futures
+            # only drops queued-but-not-started work; it cannot interrupt
+            # the thread already executing tool.execute() (see #749).
+            pool.shutdown(wait=False, cancel_futures=True)
         latency = time.time() - t0
         result.latency_seconds = latency
         result.metadata["arguments"] = params
--- a/src/openjarvis/tools/_stubs.py
+++ b/src/openjarvis/tools/_stubs.py
@@ -7,8 +7,11 @@ Each tool is registered via ``@ToolRegistry.register("name")`` and implements
 
 from __future__ import annotations
 
+import atexit
 import concurrent.futures
 import json
+import queue
+import threading
 import time
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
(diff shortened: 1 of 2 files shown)


ElliotSlusky:
Maintainer repair pushed in `8f974bba`: tool execution now uses a bounded shared daemon runner (8 running + 8 queued), rejects excess work with clear backpressure, cancels queued timed-out futures, and shuts down without waiting on stuck calls. Added repeated-timeout/backpressure and subprocess-exit regressions. Verified: focused timeout suite (9 passed), broader executor suites (33 passed), plus Ruff check/format.

ElliotSlusky (approved):

PR #795, opened 2026-08-17
fix(mcp): drain stdio transport stderr to prevent deadlock

### LINKED ISSUE #750: Stdio MCP server writing to stderr can deadlock the whole process
Description: `StdioTransport` spawns the stdio MCP server with `stderr=subprocess.PIPE` (src/openjarvis/mcp/transport.py:70-74) but nothing in the package ever reads from that pipe. Once the child process writes more than the OS pipe buffer to stderr (a few KB), it blocks on its own write and never answers on stdout, so `proc.stdout.readline()` (transport.py:86), which has no timeout, blocks forever. Because `MCPClient.initialize()` runs synchronously on the `jarvis ask` / `jarvis serve` startup path, a single chatty server (e.g. one `npx` first-run banner) wedges the whole process before it ever prints anything.

```python
def _start(self) -> None:
    self._process = subprocess.Popen(
        self._command,
        stdin=subprocess.PIPE,
        stdout=subprocess.PIPE,
        stderr=subprocess.PIPE,
        text=True,
    )

def send(self, request: MCPRequest) -> MCPResponse:
    ...
    response_line = proc.stdout.readline()
```

Steps to Reproduce:
1. Configure an MCP stdio server whose process, on startup, writes more than a few KB to stderr before answering any JSON-RPC request on stdout (a small test script works: print 2000+ lines to stderr, then read one line from stdin and echo a JSON-RPC response).
2. Call `StdioTransport(command=[...]).send(request)`.
3. Observe the call never returns - the process is blocked writing to its own stderr pipe, and the read on stdout hangs indefinitely.

Expected Behavior: The transport works regardless of how much the child writes to stderr, since nothing intentionally waits on it.

Actual Behavior: `send()` hangs forever once the child's stderr output exceeds the OS pipe buffer.

Environment: OpenJarvis @ 1ca3ec6 (main), Python 3.13.1, Windows 11. Platform-independent (the exact pipe buffer size varies by OS, but the deadlock exists everywhere `subprocess.PIPE` is used for stderr with nothing reading it).

Suggested fix: Start a background thread that continuously drains the child's stderr (e.g. mirroring each line to a logger and keeping a small rolling tail for diagnostics), and join that thread on `close()`. A local patch doing this, plus a regression test, is available if useful.

havolline1:
## Summary
- Fixes #750: `StdioTransport` spawns its subprocess with `stderr=subprocess.PIPE` but nothing ever reads from that pipe. Once the child writes more than the OS pipe buffer to stderr (a few KB), the child blocks on its own write and never gets to read stdin or answer on stdout — deadlocking `proc.stdout.readline()` in `send()` forever.
- Adds a daemon background thread that continuously drains `proc.stderr` (logging each line at debug level) for the lifetime of the transport, started in `_start()` and joined in `close()`.

## Test plan
- [x] Added `test_survives_large_stderr_output` (`tests/mcp/test_transport.py`): spawns a child that writes 20,000 lines to stderr before touching stdin, then calls `send()` from a worker thread with a join timeout. Confirmed it deadlocks (times out) against the old code and passes after the fix.
- [x] `uv run pytest tests/mcp/` — 149 passed
- [x] `uv run pytest tests/security/test_tool_confirmation.py tests/security/test_security_wiring.py tests/security/test_boundary_guard.py` — 30 passed
- [x] Ran the full suite with the same extras/markers as CI (`uv sync --extra dev --extra framework-comparison --extra server`, `pytest tests/ -n auto -m "not live and not cloud and not hub"`). 162 pre-existing failures unrelated to this change (missing compiled `openjarvis_rust` extension, Linux-only RAPL sysfs paths invalid on this Windows box, etc.) — verified identical on a clean `main` checkout via `git stash`, so none are caused by this PR.

### DIFF
--- a/src/openjarvis/mcp/transport.py
+++ b/src/openjarvis/mcp/transport.py
@@ -2,7 +2,9 @@
 
 from __future__ import annotations
 
+import logging
 import subprocess
+import threading
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Any, List, Optional
 
@@ -10,6 +12,8 @@ from openjarvis.mcp.protocol import MCPRequest, MCPResponse
 
 if TYPE_CHECKING:
     from openjarvis.mcp.server import MCPServer
+
+logger = logging.getLogger(__name__)
 
 
 class MCPTransport(ABC):
@@ -61,6 +65,7 @@ class StdioTransport(MCPTransport):
     def __init__(self, command: List[str]) -> None:
         self._command = command
         self._process: Optional[subprocess.Popen[str]] = None
+        self._stderr_thread: Optional[threading.Thread] = None
         self._start()
 
     def _start(self) -> None:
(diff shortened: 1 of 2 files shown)


ElliotSlusky (approved):
Maintainer review at the current head: approved. Focused MCP transport tests: 18 passed; Ruff passed. The stderr drain and lifecycle handling fix the deadlock without a blocking regression. Repository rules still require a code-owner approval before merge.

PR #796, opened 2026-08-17
fix(server): exempt OPTIONS requests from API key auth

### LINKED ISSUE #758: Enabling OPENJARVIS_API_KEY breaks CORS preflight for cross-origin clients
Description: `app.py` registers `CORSMiddleware` early (around line 211) and `AuthMiddleware` last (around line 456); since Starlette's `add_middleware` makes the most-recently-added middleware outermost, the effective order is `AuthMiddleware -> SecurityHeaders -> CORSMiddleware -> routes`. `AuthMiddleware.dispatch` has no exemption for `OPTIONS` requests (src/openjarvis/server/auth_middleware.py:41), so every CORS preflight (an `OPTIONS` request with no `Authorization` header, which browsers never send on a preflight) hits `AuthMiddleware` first and gets a bare 401 that never passes through `CORSMiddleware` - so it carries no `Access-Control-Allow-Origin` header either. Enabling `OPENJARVIS_API_KEY`, the documented hardening step, therefore breaks the Tauri desktop app and any cross-origin browser client at the network layer, while same-origin tools like `curl` keep working.

```python
async def dispatch(self, request: Request, call_next):
    if self._api_key and self._requires_auth(request.url.path):
        auth = request.headers.get("Authorization", "")
        if not auth:
            return JSONResponse({"detail": "Missing Authorization header"}, status_code=401)
        ...
    return await call_next(request)
```

Steps to Reproduce:
1. Start the server with `OPENJARVIS_API_KEY` set to a non-empty value.
2. From a different origin (e.g. a browser page served from a different port, or `fetch()` with `mode: "cors"`), send an `OPTIONS` preflight to a protected route like `/v1/chat/completions`.
3. Observe the response is a 401 with no `Access-Control-Allow-Origin` header, so the browser reports an opaque CORS/network error rather than surfacing the 401 to application code.

Expected Behavior: `OPTIONS` preflight requests bypass authentication (mirroring the existing exemption in `SecurityHeadersMiddleware`), and `CORSMiddleware` is the outermost layer so every response - including 401s - carries the correct CORS headers.

Actual Behavior: Preflights are rejected by `AuthMiddleware` before `CORSMiddleware` runs, breaking cross-origin access entirely once an API key is configured.

Suggested fix: Short-circuit `AuthMiddleware.dispatch` for `request.method == "OPTIONS"`, and register `CORSMiddleware` after `SecurityHeadersMiddleware`/`AuthMiddleware` so it ends up outermost. A local patch doing both, plus regression tests, is available if useful.

havolline1:
## Summary
- Fixes #758: `AuthMiddleware` is registered after `CORSMiddleware` in `create_app()`, and Starlette applies middleware in reverse registration order, so `AuthMiddleware` sits outermost (`AuthMiddleware -> SecurityHeaders -> CORSMiddleware -> routes`). `dispatch()` had no exemption for `OPTIONS`, so browser CORS preflights — which never carry an `Authorization` header — were rejected with a bare 401 before ever reaching `CORSMiddleware`. No `Access-Control-Allow-*` headers ever made it onto the response, so the browser treated the whole cross-origin request as blocked whenever `OPENJARVIS_API_KEY` was set.
- Fix: `AuthMiddleware.dispatch()` now passes `OPTIONS` requests straight through via `call_next`, letting `CORSMiddleware` (further in) handle and answer the preflight. Preflights carry no credentials to protect, so there's nothing to authenticate.
- Scope note: I kept this to the minimal root-cause fix (the missing OPTIONS exemption) rather than also reordering `CORSMiddleware` to be outermost, since `SecurityHeaders` already skips OPTIONS (existing test `test_middleware_skips_options`) and the added tests confirm CORS headers now reach the client correctly with the current ordering.

## Test plan
- [x] Added `test_options_preflight_exempt` (`tests/server/test_auth_middleware.py`) — a bare `AuthMiddleware` must not 401 an OPTIONS request.
- [x] Added `tests/server/test_cors_auth_interaction.py` — full `create_app()` integration test: an OPTIONS preflight to `/v1/models` with `OPENJARVIS_API_KEY` set gets `Access-Control-Allow-Origin` back (not 401), while an actual unauthenticated GET to the same route still correctly 401s (non-regression check).
- [x] Confirmed both new tests fail against the old code (401 instead of a CORS-handled response) and pass after the fix.
- [x] `uv run pytest tests/server/ tests/security/test_network_defaults.py` — 371 passed

### DIFF
--- a/src/openjarvis/server/app.py
+++ b/src/openjarvis/server/app.py
@@ -208,14 +208,6 @@ def create_app(
             "https://tauri.localhost",
         ]
     )
-    app.add_middleware(
-        CORSMiddleware,
-        allow_origins=_origins,
-        allow_credentials=True,
-        allow_methods=["*"],
-        allow_headers=["*"],
-    )
-
     # Store dependencies in app state
     app.state.engine = engine
     app.state.model = model
@@ -456,6 +448,18 @@ def create_app(
             app.add_middleware(AuthMiddleware, api_key=api_key)
         except Exception as exc:
             logger.debug("Auth middleware init skipped: %s", exc)
+
+    # Register CORS last so it is the outermost middleware. In addition to
+    # handling preflights, this ensures browser clients can read 401 responses
+    # produced directly by AuthMiddleware instead of seeing an opaque CORS
+    # network error.
+    app.add_middleware(
+        CORSMiddleware,
+        allow_origins=_origins,
+        allow_credentials=True,
+        allow_methods=["*"],
+        allow_headers=["*"],
+    )
 
     # Mount webhook routes (always — SendBlue may be configured dynamically)
     if webhook_config:
--- a/src/openjarvis/server/auth_middleware.py
+++ b/src/openjarvis/server/auth_middleware.py
@@ -42,8 +42,9 @@ class AuthMiddleware(BaseHTTPMiddleware):
         # Browser CORS preflights never carry Authorization, and rejecting
         # them here means they never reach CORSMiddleware to get
         # Access-Control-Allow-* headers, breaking cross-origin access
-        # outright (#758). Preflights carry no credentials to protect.
-        if request.method == "OPTIONS":
+        # outright (#758). Do not exempt arbitrary OPTIONS requests: a real
+        # preflight is identified by both headers required by the CORS protocol.
+        if self._is_cors_preflight(request):
             return await call_next(request)
         if self._api_key and self._requires_auth(request.url.path):
             auth = request.headers.get("Authorization", "")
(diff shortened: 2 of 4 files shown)


ElliotSlusky:
Maintainer repair pushed in `15d7eb3b`: the auth exemption now applies only to real CORS preflights (`OPTIONS` with both required headers), and CORS is outermost so allowed origins receive CORS headers even on auth failures. Added partial/plain OPTIONS, allowed/disallowed origin, preflight, and 401 coverage. Verified: relevant server suites (21 passed), plus Ruff check/format.

ElliotSlusky (approved):

PR #797, opened 2026-08-17
fix(memory): cross-process locking for LocalFactStore

### LINKED ISSUE #755: LocalFactStore corrupts memory_facts.jsonl when two processes write concurrently
Description: `LocalFactStore.add/list/clear/count` do a read-modify-write over the shared JSONL file (`_sync_from_disk_locked` -> dedupe -> append -> `_flush()` full rewrite), guarded only by a `threading.Lock`, which has no effect across processes (src/openjarvis/memory/store.py:86-172). Two processes that share the default `~/.openjarvis/memory_facts.jsonl` - e.g. `jarvis serve` and a concurrently running `jarvis chat`, both starting a `MemoryService` - can clobber each other, since the second `_flush()` rewrites the whole file from a snapshot taken before the other process's write. `_flush()` also uses a fixed temp filename (`memory_facts.jsonl.tmp`), so two concurrent flushes can truncate and interleave into the same temp file before either `os.replace()` runs.

```python
def __init__(self, path=None, *, max_facts: int = 1000) -> None:
    ...
    self._lock = threading.Lock()
    self._facts: List[Fact] = self._load()

def _flush(self) -> None:
    tmp = self._path.with_suffix(self._path.suffix + ".tmp")
    ...
    tmp.write_text(payload, encoding="utf-8")
    os.replace(tmp, self._path)
```

Steps to Reproduce:
1. Create two `LocalFactStore` instances pointed at the same path, simulating two processes (e.g. two separate Python interpreters, or two threads that each construct their own store instance without sharing the in-process lock).
2. Have store A begin `add()` (load, dedupe) and, before it flushes, have store B complete its own `add()` and flush.
3. Let store A finish and flush.
4. Observe only store A's fact ends up on disk - store B's addition is lost.

Expected Behavior: Facts added by either process/store instance survive, or the store makes a documented, deliberate choice about concurrent-writer semantics.

Actual Behavior: The class docstring claims cross-caller safety, but concurrent writers from different processes silently lose data.

Environment: OpenJarvis @ 1ca3ec6 (main), Python 3.13.1, Windows 11. Platform-independent - the class docstring's cross-process safety claim applies to any OS running two OpenJarvis processes against the same config directory.

Suggested fix: Add a cross-process file lock (e.g. `fcntl.flock` on POSIX, `msvcrt.locking` on Windows) around the critical section of `add`/`list`/`clear`/`count`, and use a unique per-call temp filename (e.g. via `tempfile.mkstemp`) for `_flush()` so two concurrent flushes cannot interleave. A local patch implementing both, plus regression tests, is available if useful.

havolline1:
## Summary
- Fixes #755: `LocalFactStore` only used a `threading.Lock()`, which has no effect across processes. The read-modify-write cycle (load -> dedupe -> append -> flush) let two processes race: process A loads and starts modifying, process B finishes its own modification and rewrites the whole file, then A flushes its now-stale snapshot and clobbers B's write. On top of that, both processes wrote to the same fixed temp filename (`memory_facts.jsonl.tmp`), so concurrent flushes could interleave into the same temp file before either `os.replace()` ran — the regression test below reproduced this as an outright `FileNotFoundError` on the losing side.
- `add()` and `clear()` now hold a blocking, cross-process exclusive lock (`fcntl.flock` on POSIX, `msvcrt.locking` on Windows, via a new `_cross_process_lock()` context manager keyed on a sibling `.lock` file) around the entire load -> modify -> flush cycle. `list()`/`count()` are left unlocked at the cross-process level since `os.replace()` is atomic — a reader only ever sees a complete old or new file, never a torn one.
- `_flush()` now uses `tempfile.mkstemp()` for a unique temp filename per flush instead of a fixed `<path>.tmp`, so concurrent flushes can never step on each other's temp file even before the new lock closes that window.

## Test plan
- [x] Added `test_concurrent_instances_do_not_lose_writes` (`tests/memory/test_fact_store.py`): 8 threads, each owning its own `LocalFactStore` instance pointed at the same file (simulating separate processes — the per-instance `threading.Lock` can't coordinate between them), each adding 15 facts concurrently; asserts all 120 survive.
- [x] Confirmed the test fails reliably against the old code — deterministically reproduced the *exact* failure mode from the issue (shared temp filename collision -> `FileNotFoundError` when the losing side's renamed-away temp file vanishes), 3/3 runs.
- [x] Confirmed the test passes reliably with the fix, 5/5 runs.
- [x] `uv run pytest tests/memory/test_fact_store.py` — 17 passed
- [x] `uv run pytest tests/memory/ tests/cli/test_memory_cmd.py`, excluding modules that need the compiled `openjarvis_rust` extension (not built in this environment, pre-existing/unrelated) — 100 passed

### DIFF
--- a/src/openjarvis/memory/store.py
+++ b/src/openjarvis/memory/store.py
@@ -11,20 +11,64 @@ from __future__ import annotations
 
 import json
 import os
+import sys
+import tempfile
 import threading
 import time
 from abc import ABC, abstractmethod
+from contextlib import contextmanager
 from dataclasses import asdict, dataclass
 from pathlib import Path
-from typing import Any, Iterable, List
+from typing import Any, Iterable, Iterator, List
 
 from openjarvis.core.paths import get_config_dir
 from openjarvis.core.registry import FactStoreRegistry
+
+if sys.platform == "win32":
+    import msvcrt
+else:
+    import fcntl
 
 
 def _default_fact_path() -> Path:
     """Return the env-aware default JSONL path for automatic memory facts."""
     return get_config_dir() / "memory_facts.jsonl"
+
+
+@contextmanager
+def _cross_process_lock(lock_path: Path) -> Iterator[None]:
+    """Hold an exclusive, blocking OS-level lock on *lock_path*.
+
+    ``threading.Lock`` only serializes within one process, so it cannot
+    protect the load -> modify -> flush cycle when two separate
+    ``LocalFactStore`` instances (e.g. in different processes) touch the
+    same file (#755). This blocks until the lock is free rather than
+    failing fast, since callers want to wait their turn, not error out.
+    """
+    lock_path.parent.mkdir(parents=True, exist_ok=True)
+    f = open(lock_path, "a+b")
+    try:
+        if sys.platform == "win32":
+            # msvcrt locks a byte range; the file needs at least one byte
+            # for the region at offset 0 to be lockable.
+            f.seek(0, os.SEEK_END)
+            if f.tell() == 0:
+                f.write(b"\0")
+                f.flush()
+            f.seek(0)
+            msvcrt.locking(f.fileno(), msvcrt.LK_LOCK, 1)
+        else:
+            fcntl.flock(f.fileno(), fcntl.LOCK_EX)
+        try:
+            yield
+        finally:
+            if sys.platform == "win32":
+                f.seek(0)
+                msvcrt.locking(f.fileno(), msvcrt.LK_UNLCK, 1)
+            else:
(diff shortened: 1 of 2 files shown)


ElliotSlusky:
Maintainer repair pushed in `6ee42937`: Windows locking now retries nonblocking CRT lock contention until acquisition instead of relying on the CRT fixed retry limit, while unexpected errors still propagate. Added real spawned-process write and add/clear race coverage plus simulated Windows contention/error tests. Verified: focused fact-store suite (21 passed), full memory suite (161 passed, 14 skipped), plus Ruff check/format.

ElliotSlusky (approved):

Issue #752, opened 2026-08-17
A stray non-JSON line on stdout drops an MCP server for the session


Lupin133:
Description: `StdioTransport.send()` hands the first stdout line straight to `MCPResponse.from_json` (transport.py:89), which raises `json.JSONDecodeError` if that line isn't valid JSON. Node MCP servers launched via `npx -y ...` - exactly what `jarvis add` generates - commonly print npm notices to stdout on a cold cache or version mismatch, so the loader logs a generic `Failed to discover MCP tools from 'x': Expecting value: line 1 column 1 (char 0)` and drops the server for the entire session, with no indication of which line caused it.

```python
response_line = proc.stdout.readline()
if not response_line:
    raise RuntimeError("No response from subprocess")
return MCPResponse.from_json(response_line.strip())
```

Steps to Reproduce:
1. Write a fake stdio MCP server that prints `npm warn exec The following package was not found and will be installed` to stdout, then answers subsequent JSON-RPC requests correctly.
2. Point `MCPClient` at it and call `list_tools()`.
3. Observe `json.JSONDecodeError` and zero tools discovered, even though the server would have answered correctly afterward.

Expected Behavior: A stray non-JSON banner line on stdout is skipped (or at minimum logged with enough context to diagnose), and discovery proceeds using the real JSON-RPC response.

Actual Behavior: The first non-JSON line aborts discovery for that server with an opaque error.

Environment: OpenJarvis @ 1ca3ec6 (main), Python 3.13.1, Windows 11. Platform-independent (the `npx` cold-cache banner is a common trigger on any OS).

Suggested fix: In the same read loop that fixes response-id correlation, attempt `json.loads` per line and skip (logging at a lower level) any line that fails to parse, up to a bounded budget, instead of failing on the first line encountered. A local patch implementing this, plus a regression test, is available if useful.

ElliotSlusky:
Closing as fixed by the merged MCP stdio transport work. Current main uses a bounded stdout reader, correlates response IDs, skips notifications/stale/non-response lines defensively, drains stderr, serializes requests, and applies a monotonic response deadline. This prevents one stray line from dropping the MCP session.

PR #798, opened 2026-08-17
fix(mcp): validate response id in StdioTransport.send()

### LINKED ISSUE #751: MCP client treats any stdout line as the response with no id check
Description: `StdioTransport.send()` writes the request and reads exactly one line from stdout, returning whatever it parses as the response (src/openjarvis/mcp/transport.py:86-89), and `MCPClient` never compares the response's `id` to the request's `id`. MCP servers may legitimately push unsolicited messages on stdout (`notifications/message`, `notifications/progress`, ...); when one of those arrives before the real response, `MCPResponse.from_json` parses it into `result=None, error=None`, and `MCPClient.initialize()`/`list_tools()` then call `.get()` on that `None`, raising `AttributeError`. Because that error is per-server and swallowed by the loader's blanket `except Exception`, the server is simply dropped with a generic message; worse, the real response is left unread in the pipe, so every later request on that connection reads the previous request's answer.

```python
def send(self, request: MCPRequest) -> MCPResponse:
    ...
    response_line = proc.stdout.readline()
    if not response_line:
        raise RuntimeError("No response from subprocess")
    return MCPResponse.from_json(response_line.strip())
```
```python
response = self._send("initialize", params)
self._initialized = True
self._capabilities = response.result.get("capabilities", {})  # AttributeError if result is None
```

Steps to Reproduce:
1. Write a fake stdio MCP server that first emits a `notifications/message` line, then answers a `tools/list` request correctly with a matching `id`.
2. Point `MCPClient` at it and call `list_tools()`.
3. Observe `AttributeError: 'NoneType' object has no attribute 'get'` (surfaced by loader.py as "Failed to discover MCP tools from '<name>'").

Expected Behavior: The client skips unsolicited server messages that lack a matching request id, and returns the correctly-correlated response.

Actual Behavior: The first line on stdout is treated as the response unconditionally; a notification arriving out of band crashes discovery and desynchronizes every subsequent request on that connection.

Suggested fix: Have `send()` loop over incoming lines, skip lines with no `id` (notifications) and lines whose `id` doesn't match the outstanding request, and return the first correlated response (bounded by a max-lines-skipped budget to avoid an infinite loop). Have `MCPClient` raise an actionable error when a response carries neither `result` nor `error`. A local patch implementing both, plus regression tests, is available if useful.

havolline1:
## Summary
- Fixes #751 **and** #752 — both describe the same root cause in `StdioTransport.send()`: it read exactly one line from stdout and handed it straight to `MCPResponse.from_json()`, with no check that the line was actually JSON, actually a response (vs. a notification), or actually addressed to the request just sent.
  - **#751**: an unsolicited server notification (no `id`, no `result`/`error`) arriving before the real response parsed as a bogus `MCPResponse(result=None, error=None)` instead of erroring, leaving the real response unread and desyncing every later call.
  - **#752**: a stray non-JSON line (e.g. an npm warning banner from a server launched via `npx -y` on a cold cache) raised `json.JSONDecodeError` straight out of `send()`, which the loader's per-server isolation (`loader.py`) caught and used to silently drop that server for the whole session — even though it would have answered correctly moments later.
- `send()` now loops, reading lines until it finds a well-formed JSON-RPC object with `id` matching the outstanding request **and** either `result` or `error` present. Non-JSON lines, notifications (no `id`), and stale/mismatched-id responses are all skipped rather than accepted or raised.
- Bounded by a `_MAX_SKIPPED_LINES` budget (1000) so a server that spams noise or never answers raises a clear `RuntimeError` instead of hanging `send()` forever.

## Test plan
- [x] `test_skips_unsolicited_notification_before_response` (#751): a fake server emits a `notifications/progress` line (no id) before the real response; asserts the client returns the correct response and stays in sync for a following call.
- [x] `test_skips_response_with_stale_id` (#751): a fake server emits a response with a mismatched id before the real one; asserts the client waits for the matching id.
- [x] `test_gives_up_after_max_skipped_lines` (#751): a fake server floods stdout with 2000 non-matching notifications and never answers; asserts `send()` raises `RuntimeError` instead of hanging.
- [x] `test_skips_non_json_banner_before_response` (#752): a fake server prints an npm-style warning banner (non-JSON) before answering; asserts discovery-style calls succeed instead of raising `json.JSONDecodeError`. Confirmed this reproduces the exact `JSONDecodeError` from the issue against the old single-readline code, and passes with the fix.
- [x] Confirmed all four fail against the old code and pass after the fix.
- [x] `uv run pytest tests/mcp/` — 152 passed

## Note on #750
This PR was branched directly off `main` and does not depend on #795 (the stdio stderr-deadlock fix, also currently open) — the diff here touches only `send()`, while #795 touches `_start()`/`close()`. They apply independently in either merge order.


ElliotSlusky:
Maintainer repair pushed in `d558e117`: stdio reads now run through one bounded, cross-platform reader queue with a real monotonic response deadline; blank/noise floods cannot extend it, and large valid notification bursts no longer hit an arbitrary line cap before the matching response. Close wakes readers and force-kills an uncooperative child after the grace period. Verified: transport suite (24 passed), full MCP suite (155 passed), plus Ruff check/format.

ElliotSlusky (dismissed):

PR #799, opened 2026-08-18
fix(server): serialize per-request model overrides on shared agents

### LINKED ISSUE #759: Concurrent chat requests with different models race on the shared agent
Description: `_handle_agent` temporarily overwrites the shared `app.state.agent`'s `_model` attribute to honor a per-request `model` override, then restores it in a `finally` (src/openjarvis/server/routes.py:544-556). Both the non-streaming and streaming chat paths run `_handle_agent` via `asyncio.to_thread`, so concurrent requests carrying different `model` values race on the same mutable attribute with no lock: request B can read A's override as its own "original" value, and whichever request's `finally` runs last permanently pins the shared agent to the wrong model for every later request that omits `model`.

```python
original_model = agent._model
if model:
    agent._model = model
try:
    ...
    result = agent.run(input_text, context=ctx)
finally:
    agent._model = original_model
```

Steps to Reproduce:
1. Start the server with a single shared agent (`app.state.agent`).
2. Fire two concurrent `POST /v1/chat/completions` requests specifying two different `model` values, timed to overlap (e.g. with a mocked agent whose `run()` sleeps briefly).
3. Fire a third request with no `model` field afterward.
4. Observe the third request runs against whichever of the two override values was restored last, not the agent's originally configured default.

Expected Behavior: Concurrent requests with different `model` overrides do not interfere with each other or leave the shared agent pinned to an unrelated model.

Actual Behavior: `agent._model` is mutated with no synchronization, so overlapping requests can corrupt the shared, restored value.

Suggested fix: Serialize the override/run/restore critical section per agent instance with a lock (e.g. keyed off the agent object), so concurrent overrides on the same shared agent no longer interleave. A local patch doing this, plus a regression test, is available if useful.

havolline1:
## Summary
- Fixes #759: `_handle_agent` temporarily overwrites the shared `agent._model` for the duration of one request (`original = agent._model; agent._model = model; try: agent.run(...); finally: agent._model = original`), with no synchronization. Both the streaming and non-streaming chat routes run this via `asyncio.to_thread`, so concurrent requests with different `model` values execute on real OS threads and can genuinely interleave: request B can read request A's still-live override as its own "original" value, and whichever request's cleanup runs last permanently pins the shared agent to the wrong model — corrupting every subsequent request, including ones with no override at all.
- Adds a per-agent-instance lock (`_get_agent_model_lock`, backed by a `WeakKeyDictionary` so unrelated agent instances never serialize against each other) around the **entire** override-run-restore cycle, not just the assignment lines — `agent.run()` itself reads `self._model` throughout the call, so the critical section has to cover the whole thing.

## Test plan
- [x] Added `test_concurrent_model_overrides_do_not_corrupt_shared_agent` (`tests/server/test_agent_model_override.py`): 16 threads call `_handle_agent` concurrently on one shared fake agent, each with a distinct `model` override; the fake agent's `run()` sleeps on either side of reading `self._model` to widen the race window. Asserts every worker observed its *own* model during its own call (no cross-contamination) and that the agent is back to its true default once all requests complete.
- [x] Confirmed the test fails against the old code — reproduced the exact cross-contamination described in the issue (one worker observed a completely different worker's override) — and passes reliably with the fix.
- [x] `uv run pytest tests/server/` — 359 passed

### DIFF
--- a/src/openjarvis/server/routes.py
+++ b/src/openjarvis/server/routes.py
@@ -519,8 +519,12 @@ def _handle_direct(
 # interleave. A lock per agent *instance* serializes the
 # override-run-restore critical section below so one request's temporary
 # `model` override can never be read as another's "original" value (#759).
-_agent_model_locks: "weakref.WeakKeyDictionary[Any, threading.Lock]" = (
-    weakref.WeakKeyDictionary()
+# Key by object identity rather than by the agent itself: custom agents may be
+# unhashable or may not support weak references.  The values are weak so this
+# registry retains neither the agent nor an idle lock.  A caller keeps the lock
+# alive from lookup through the full override/run/restore critical section.
+_agent_model_locks: "weakref.WeakValueDictionary[int, threading.Lock]" = (
+    weakref.WeakValueDictionary()
 )
 _agent_model_locks_guard = threading.Lock()
 
@@ -529,11 +533,12 @@ def _get_agent_model_lock(agent: Any) -> threading.Lock:
     """Return the lock serializing model overrides for *agent*, creating it
     on first use. One lock per agent instance, not global, so unrelated
     agents don't serialize against each other."""
+    agent_id = id(agent)
     with _agent_model_locks_guard:
-        lock = _agent_model_locks.get(agent)
+        lock = _agent_model_locks.get(agent_id)
         if lock is None:
             lock = threading.Lock()
-            _agent_model_locks[agent] = lock
+            _agent_model_locks[agent_id] = lock
         return lock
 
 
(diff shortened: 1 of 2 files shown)


ElliotSlusky:
Pushed follow-up commit 7062a61. The lock registry is now keyed by object identity with weak lock values, so unhashable and non-weakrefable custom agents retain per-instance serialization without retaining agents or idle locks. Verification: pytest -q tests/server/test_agent_model_override.py (3 passed), plus Ruff check and format check.

ElliotSlusky (approved):

PR #801, opened 2026-08-18
fix(memory): enable WAL mode and longer busy_timeout for SQLite backend

### LINKED ISSUE #756: SQLite memory backend has no busy_timeout or WAL, contention fails fast
Description: `SQLiteMemory::new` opens the connection with `Connection::open` and never sets `journal_mode` or `busy_timeout` (rust/crates/openjarvis-tools/src/storage/sqlite.rs:37). Without WAL mode, `memory.db` runs in rollback-journal mode where any concurrent writer blocks all readers, and rusqlite's default 5-second busy handler turns ordinary contention into a hard `database is locked` error rather than a bounded wait. Multiple connections to the same file are the norm in this codebase (`cli/ask.py`, `serve.py`, `api_routes.py` each open their own), each guarded only by its own in-process mutex, so cross-connection contention is common.

```rust
let conn = Connection::open(db_path).map_err(|e| {
    OpenJarvisError::Io(std::io::Error::other(e.to_string()))
})?;

conn.execute_batch(
    "CREATE TABLE IF NOT EXISTS documents ( ... );
     CREATE VIRTUAL TABLE IF NOT EXISTS documents_fts USING fts5( ... );",
)
```

Steps to Reproduce:
1. Open two `SQLiteMemory` connections against the same `memory.db` file.
2. On one connection, begin (and hold open) a write transaction.
3. On the other connection, call `store()`.
4. Observe `PRAGMA journal_mode` reports `delete` (not `wal`), and the second `store()` call fails after about 5 seconds with `IO error: database is locked` instead of waiting for the first transaction to finish.

Expected Behavior: Concurrent access to the same memory database is either serialized transparently (WAL mode, readers not blocked by a writer) or waits a reasonable, configurable amount of time before failing.

Actual Behavior: Readers are blocked by any writer, and writers fail after only 5 seconds of contention.

Environment: OpenJarvis @ 1ca3ec6 (main), Python 3.13.1 (via the Rust extension), Windows 11. Platform-independent. Note: `SQLiteMemory::new`'s handling of a leading `~` in the db path resolves it via the `HOME` environment variable, which is not guaranteed to be set on Windows (Windows conventionally uses `USERPROFILE`); this is a separate, narrower issue not addressed by the fix below and is called out here only as a platform-specific caveat.

Suggested fix: Immediately after `Connection::open`, set a longer `busy_timeout` (e.g. 10s) and switch to `PRAGMA journal_mode=WAL`, tolerating the pragma being a no-op for `:memory:` databases or filesystems that refuse WAL. A local patch doing this, plus a Rust-level regression test, is available if useful.

havolline1:
## ⚠️ Unverified locally — no Rust toolchain in this environment

Unlike my other PRs on this repo, I could not compile or run `cargo test`/`cargo clippy` for this change — this dev environment has no `rustc`/`cargo` installed at all. I've written this fix and its test carefully against the existing rusqlite 0.32 API and this crate's own conventions (see below), but **CI's `rust` job is the first real compile/test signal for this PR.** Opening as a draft for that reason — please have CI validate before review, and I'm glad to iterate on any compile/clippy failures it surfaces.

## Summary
- Fixes #756: `SQLiteMemory::new()` opened its `Connection` with no `journal_mode` or `busy_timeout` configured. Without WAL mode, `memory.db` runs in rollback-journal mode, where any concurrent writer blocks all readers; combined with rusqlite's default 5s `busy_timeout` (too short under realistic contention), this surfaces as "database is locked" errors instead of transparent serialization.
- Right after `Connection::open`, now sets a 10s `busy_timeout` (via `Connection::busy_timeout`) and `PRAGMA journal_mode=WAL;` (via `execute_batch`, matching the existing convention in `rust/crates/openjarvis-learning/src/optimize/store.rs`, which already does this for its own SQLite connections).
- WAL is a silent no-op for in-memory databases and some filesystems (e.g. network shares) — SQLite just keeps the existing journal mode there, same as the precedent this mirrors.

## Test plan (unverified — needs CI)
- Added `test_sqlite_uses_wal_journal_mode_and_longer_busy_timeout` to the crate's existing `#[cfg(test)] mod tests`: opens a file-backed `SQLiteMemory` (via `tempfile::tempdir`, already a dev-dependency of this crate) and asserts `PRAGMA journal_mode` reports `wal` and `PRAGMA busy_timeout` reports `>= 10000` ms.
- Not run locally. CI's `rust` job runs `cargo clippy --workspace --all-targets -- -D warnings` then `cargo test --workspace`, which will validate this.


ElliotSlusky:
Pushed follow-up commit 8b3be27 and kept this PR in draft. WAL activation is now best-effort on unsupported SQLite targets, while the 10 second busy timeout remains enforced. Added contention coverage for readers during an active writer and serialization of competing writers. Local cargo test -p openjarvis-tools sqlite reached dependency compilation but the host linker was blocked by an unaccepted Xcode license before project code compiled; hosted Rust CI is the verification gate.

ElliotSlusky:
Pushed follow-up commit 7c1ab4b to fix the Rust CI compile failure in the new contention tests. The scoped workers now move their channel senders while borrowing the SQLite handles, preserving post-scope assertions. Local `cargo clippy -p openjarvis-tools --all-targets -- -D warnings` remains blocked before project compilation by the host’s unaccepted Xcode license; hosted Rust CI will validate the fix.

ElliotSlusky:
Hosted CI run 32318721367 is fully green on 7c1ab4b: Rust Clippy + workspace tests, Linux Python tests, lint, and Windows 3.12/3.13 all passed.

ElliotSlusky (dismissed):

PR #802, opened 2026-08-18
fix(memory): reject non-positive max_facts instead of disabling the cap

### LINKED ISSUE #757: LocalFactStore(max_facts=0) disables the cap instead of storing nothing
Description: `LocalFactStore.__init__` clamps with `self._max_facts = max(0, int(max_facts))` (src/openjarvis/memory/store.py:86), and the eviction check is `if self._max_facts and len(self._facts) > self._max_facts:` (store.py:147). Since `0` (and any negative value, clamped to `0`) is falsy, this disables the cap entirely instead of enforcing it. The documented contract (`core/config.py`, `max_facts: int = 1000  # cap on stored facts (oldest evicted past the cap)`) describes a bound with no mention of an "unlimited" escape value, so a user who sets `max_facts = 0` expecting to keep nothing (or a strict minimum) instead gets unbounded growth of `memory_facts.jsonl`, which is fully rewritten and deduped on every added fact.

```python
self._max_facts = max(0, int(max_facts))
...
def add(self, text: str, source: str = "") -> bool:
    ...
    if self._max_facts and len(self._facts) > self._max_facts:
        self._facts = self._facts[-self._max_facts :]
```

Steps to Reproduce:
1. `store = LocalFactStore(tmp_path, max_facts=0)`
2. Call `store.add(f"fact {i}")` 50 times with distinct text.
3. Call `store.count()`.
4. Observe `count() == 50` instead of the cap being enforced (or a clear rejection of the invalid value).

Expected Behavior: A non-positive `max_facts` either enforces a cap of that size or is rejected as an invalid configuration; either way it must not silently mean "no cap".

Actual Behavior: `max_facts=0` (or any negative value) disables capping entirely, contradicting the documented "cap on stored facts" contract.

Suggested fix: Reject non-positive `max_facts` at construction with a `ValueError` explaining that it is a cap, not a switch for unlimited growth, matching the module's existing style for invalid configuration elsewhere (`create_fact_store` already raises for an unknown backend). A local patch doing this, plus regression tests, is available if useful.

havolline1:
## Summary
- Fixes #757: `LocalFactStore(max_facts=0)` silently disabled eviction entirely instead of enforcing a zero-fact cap. `if self._max_facts and len(self._facts) > self._max_facts:` short-circuits on `0` (falsy in Python), so the cap check never even ran — the opposite of "cap on stored facts" at its minimum value. Repeated `add()` calls grew the store without bound, continuously rewriting `memory_facts.jsonl`.
- `LocalFactStore.__init__` now rejects non-positive `max_facts` with `ValueError`, matching the codebase's existing error-handling convention for bad config (`create_fact_store` already raises `ValueError` for an unknown backend).
- Also simplified the eviction check now that `max_facts` is guaranteed positive: `self._facts[-self._max_facts:]` would otherwise be exposed to Python's `list[-0:]` footgun for `max_facts=0` — negative-zero doesn't exist for ints, so `-0 == 0`, and slicing `list[-0:]` returns the *whole* list rather than emptying it. That's now moot since 0 is rejected at construction, but worth calling out since a naive "just drop the truthy check" fix would have hit it.

## Test plan
- [x] Added `test_max_facts_zero_rejected` and `test_max_facts_negative_rejected` to `tests/memory/test_fact_store.py`.
- [x] Confirmed both fail against the old code (`DID NOT RAISE ValueError`) and pass after the fix.
- [x] `uv run pytest tests/memory/test_fact_store.py` — 18 passed

Branched directly off `main`; applies cleanly independent of #797 (the LocalFactStore cross-process locking fix, also currently open on the same file).

### DIFF
--- a/src/openjarvis/memory/store.py
+++ b/src/openjarvis/memory/store.py
@@ -80,10 +80,15 @@ class LocalFactStore(FactStore):
         *,
         max_facts: int = 1000,
     ) -> None:
+        max_facts = int(max_facts)
+        if max_facts <= 0:
+            raise ValueError(
+                f"max_facts must be a positive integer, got {max_facts}"
+            )
         self._path = (
             Path(path).expanduser() if path is not None else _default_fact_path()
         )
-        self._max_facts = max(0, int(max_facts))
+        self._max_facts = max_facts
         self._lock = threading.Lock()
         self._facts: List[Fact] = self._load()
 
@@ -143,8 +148,11 @@ class LocalFactStore(FactStore):
             if any(f.text.lower() == lowered for f in self._facts):
                 return False  # dedupe
             self._facts.append(Fact(text=text, source=source, created_at=time.time()))
-            # Enforce the cap by evicting the oldest entries.
-            if self._max_facts and len(self._facts) > self._max_facts:
+            # Enforce the cap by evicting the oldest entries. max_facts is
+            # always positive (validated in __init__), so this slice can't
+            # hit Python's `list[-0:]` footgun, which returns the whole
+            # list instead of emptying it.
+            if len(self._facts) > self._max_facts:
                 self._facts = self._facts[-self._max_facts :]
             self._flush()
         return True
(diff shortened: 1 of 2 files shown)


ElliotSlusky (approved):
Maintainer review at the current head: approved. Focused fact-store tests: 18 passed; Ruff passed. Non-positive cap validation is consistent with the documented bounded-store semantics. Repository rules still require a code-owner approval before merge.

PR #803, opened 2026-08-18
fix(server): repair /metrics Prometheus endpoint

### LINKED ISSUE #760: /metrics (Prometheus) endpoint always returns "no metrics available"
Description: `prometheus_metrics` calls `TelemetryAggregator.summary()`, which returns an `AggregatedStats` object - a `@dataclass(slots=True)` with fields like `total_calls` and `total_tokens` - and then accesses it with dict-style `stats.get('total_requests', 0)` (src/openjarvis/server/api_routes.py:591-618). `AggregatedStats` has no `.get` method and no `total_requests` field, so this raises `AttributeError` on the very first line building the output. The surrounding `except Exception` catches it silently, so `/metrics` returns the placeholder `# No metrics available` for every deployment, regardless of how much data has actually been recorded. Separately, unlike other routes in this file, this handler never closes the `TelemetryAggregator`'s SQLite connection (no `try/finally`), so a periodic Prometheus scrape (typically every 15s) leaks one connection per scrape even after the primary bug is fixed.

```python
agg = TelemetryAggregator(db_path)
stats = agg.summary()

lines = [
    ...
    f"openjarvis_requests_total {stats.get('total_requests', 0)}",
    ...
]
```

Steps to Reproduce:
1. Run any inference through the server so `telemetry.db` has at least one row.
2. `GET /metrics`.
3. Observe the response body is `# No metrics available` even though telemetry data exists; check server logs for `Failed to collect Prometheus metrics: 'AggregatedStats' object has no attribute 'get'`.

Expected Behavior: `/metrics` reports the real aggregated stats (request/token counts, average latency) in Prometheus text format.

Actual Behavior: Every call to `/metrics` silently fails and returns the placeholder text.

Suggested fix: Use attribute access on `AggregatedStats` (`stats.total_calls`, `stats.total_tokens`, computing average latency guarded against division by zero) instead of `.get(...)`, and wrap the aggregator use in `try/finally: agg.close()` matching the pattern used elsewhere in the same file. A local patch doing both, plus regression tests, is available if useful.

havolline1:
## Summary
- Fixes #760: `prometheus_metrics` called `stats.get('total_requests', 0)` etc. on an `AggregatedStats` *dataclass*, which has no `.get()` method. Every real request raised `AttributeError: 'AggregatedStats' object has no attribute 'get'`, silently caught by the handler's broad `except Exception` and papered over as `# No metrics available` — even with real telemetry data sitting in `telemetry.db`. There was also no `total_requests` or `avg_latency_ms` field on `AggregatedStats` to begin with (the real fields are `total_calls`, `total_tokens`, `total_latency`), and the `TelemetryAggregator`'s SQLite connection was never closed, leaking one per Prometheus scrape (typically every 15s).
- Switched to attribute access on the real fields, compute average latency as `(total_latency_seconds / total_calls) * 1000` with a guard for `total_calls == 0` (division-by-zero on an empty-but-existing `telemetry.db`), and wrapped `agg.summary()` in `try/finally: agg.close()`.

## Test plan
- [x] Added `test_metrics_endpoint_reports_real_stats`: writes a real `TelemetryRecord` via `TelemetryStore`, points `DEFAULT_CONFIG_DIR` at it, and asserts `/metrics` returns actual `openjarvis_requests_total`/`openjarvis_tokens_total`/`openjarvis_latency_avg_ms` values instead of the placeholder. Confirmed this fails against the old code with the exact `AttributeError` from the issue, and passes after the fix.
- [x] Added `test_metrics_endpoint_empty_db_no_division_by_zero`: an existing-but-empty `telemetry.db` must return `0`, not raise `ZeroDivisionError`.
- [x] `uv run pytest tests/server/test_api_routes.py` — 16 passed

Note: the pre-existing `test_metrics_endpoint` test in this file was too permissive to have ever caught this (`assert "openjarvis" in resp.text or "No metrics" in resp.text` — passes either way); left it in place since it's still a valid smoke test, but the new tests above are what actually pin the real behavior down.


ElliotSlusky (dismissed):
Maintainer review at the current head: approved. Server API route suite: 16 passed. Metrics now use AggregatedStats correctly, convert latency units safely, guard zero calls, and close the database. Repository rules still require a code-owner approval before merge.

PR #804, opened 2026-08-18
test(server): fix flaky /metrics assertion, missing the third response state

### LINKED ISSUE #792: test_metrics_endpoint only passes when the developer already has telemetry on disk
Description: `/metrics` has three response shapes — a Prometheus body when telemetry rows exist, `# No metrics available` when the store is reachable but empty, and `# no telemetry data` when there is no telemetry database at all (`src/openjarvis/server/api_routes.py`, the `DEFAULT_CONFIG_DIR` branch). The test only accepts the first two:

```python
def test_metrics_endpoint(self):
    client = TestClient(_make_app())
    resp = client.get("/metrics")
    assert resp.status_code == 200
    assert "openjarvis" in resp.text or "No metrics" in resp.text
```

Since the route resolves the database through `DEFAULT_CONFIG_DIR`, this passes on a developer machine only because it reads whatever telemetry happens to be in that developer's real `~/.openjarvis/telemetry.db`. On a machine that has never run `jarvis` — a fresh clone, a CI runner, or any run where the suite is isolated from the real home directory — the route correctly returns `# no telemetry data` and the assertion fails.

Steps to Reproduce:
1. On a machine with no `~/.openjarvis/telemetry.db` (or with `OPENJARVIS_HOME` pointed at an empty temporary directory), run:
   `pytest tests/server/test_api_routes.py -k Metrics`
2. `test_metrics_endpoint` fails with the body being `# no telemetry data`.
3. Run the same test after using `jarvis` normally at least once, and it passes.

Expected Behavior: the test passes regardless of what is in the developer's home directory, since the route legitimately answers 200 in all three states.

Actual Behavior: it fails on any machine without pre-existing telemetry — i.e. exactly on a fresh checkout.

Suggested fix: accept the third branch in the assertion (`or "no telemetry data" in resp.text`). Related: the same class of dependency exists suite-wide because tests resolve paths against the real `OPENJARVIS_HOME`; pointing it at a temporary directory in `tests/conftest.py` makes this failure reproducible and is worth doing separately.

havolline1:
## Summary
- Fixes #792: `/metrics` has three valid response states depending on telemetry data availability — Prometheus-formatted output, `"# No metrics available"` for a reachable-but-empty store, or `"# no telemetry data"` when no `telemetry.db` exists at all (the `DEFAULT_CONFIG_DIR` branch). `test_metrics_endpoint`'s assertion only covered the first two (`"openjarvis" in resp.text or "No metrics" in resp.text`) and didn't pin `DEFAULT_CONFIG_DIR` anywhere, so it depended entirely on whether the machine running the test happened to already have telemetry data under its real `~/.openjarvis` — failing on fresh installs and CI where no `telemetry.db` exists yet, and only passing after running `jarvis` normally once.
- Pinned `DEFAULT_CONFIG_DIR` to an empty `tmp_path` (deterministically exercising the "no telemetry data" state instead of leaving it up to chance) and expanded the assertion to accept all three valid states.

## Test plan
- [x] Reproduced the exact failure: with the hermetic `tmp_path` setup in place but the *old* 2-clause assertion, the test fails with `AssertionError: assert ('openjarvis' in '# no telemetry data\n' or 'No metrics' in '# no telemetry data\n')` — the precise flake described in the issue.
- [x] Confirmed it passes with the 3-clause assertion.
- [x] `uv run pytest tests/server/test_api_routes.py` — 14 passed

Branched directly off `main`; applies cleanly independent of #803 (the `/metrics` handler fix for #760, also currently open — that PR adds two more tests to this same class, but this diff doesn't touch them).

### DIFF
--- a/tests/server/test_api_routes.py
+++ b/tests/server/test_api_routes.py
@@ -120,11 +120,28 @@ class TestBudgetRoutes:
 
 
 class TestMetricsRoute:
-    def test_metrics_endpoint(self):
+    def test_metrics_endpoint(self, tmp_path, monkeypatch):
+        """Regression for #792: /metrics has three valid response states
+        (Prometheus output, "No metrics available" for a reachable-but-
+        empty store, or "no telemetry data" when no telemetry.db exists
+        at all), but the old assertion only covered two of them. Pin
+        DEFAULT_CONFIG_DIR to an empty directory so this deterministically
+        exercises the third state instead of depending on whether the
+        machine running the test happens to already have telemetry data
+        under its real ~/.openjarvis.
+        """
+        import openjarvis.core.config as config_mod
+
+        monkeypatch.setattr(config_mod, "DEFAULT_CONFIG_DIR", tmp_path)
+
         client = TestClient(_make_app())
         resp = client.get("/metrics")
         assert resp.status_code == 200
-        assert "openjarvis" in resp.text or "No metrics" in resp.text
+        assert (
+            "openjarvis" in resp.text
+            or "No metrics" in resp.text
+            or "no telemetry data" in resp.text
+        )
 
 
 class TestSkillRoutes:
--- a/tests/server/test_api_routes.py
+++ b/tests/server/test_api_routes.py
@@ -121,14 +121,11 @@ class TestBudgetRoutes:
 
 class TestMetricsRoute:
     def test_metrics_endpoint(self, tmp_path, monkeypatch):
-        """Regression for #792: /metrics has three valid response states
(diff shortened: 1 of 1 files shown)


ElliotSlusky:
Pushed follow-up commit 37a45ec. The isolated metrics test now asserts the exact no telemetry data body and text content type instead of accepting every production response state. Verification: targeted pytest passed (1 passed), plus Ruff check and format check.

ElliotSlusky (approved):

PR #805, opened 2026-08-18
fix(channels): fix Telegram send() dropping messages and false success

### LINKED ISSUE #783: Telegram send() drops replies on Markdown errors, lies on empty content
Description: `TelegramChannel.send` attaches `parse_mode="Markdown"` to every outgoing message and treats any non-2xx response as terminal (src/openjarvis/channels/telegram.py:104-147). Legacy Telegram Markdown rejects ordinary LLM output - a lone `*`, a `snake_case_identifier`, an unclosed code fence - with HTTP 400 `Bad Request: can't parse entities`, so the reply is logged and dropped with nothing surfaced to the user. Separately, `textwrap.wrap("")` returns `[]`, so `send(chat, "")` skips the send loop entirely yet still returns `True` and publishes `CHANNEL_MESSAGE_SENT` for a message that was never actually sent.

```python
chunks = textwrap.wrap(content, width=_TELEGRAM_MAX_LEN, break_long_words=True, replace_whitespace=False)
for chunk in chunks:
    payload = {"chat_id": chat_id, "text": chunk}
    if self._parse_mode:
        payload["parse_mode"] = self._parse_mode
    ...
    resp = httpx.post(url, json=payload, timeout=10.0)
    if resp.status_code >= 300:
        logger.warning("Telegram API returned status %d: %s", resp.status_code, resp.text)
        return False
self._publish_sent(channel, content, conversation_id)
return True
```

Steps to Reproduce:
1. Send a message through `TelegramChannel.send` whose text contains a lone `*` or an unclosed code fence - content a real LLM commonly produces.
2. Observe the Telegram API returns HTTP 400 `can't parse entities`, `send()` returns `False`, and the user never receives the reply (with no fallback attempt).
3. Separately, call `send(chat_id, "")`.
4. Observe `send()` returns `True` and a `CHANNEL_MESSAGE_SENT` event is published, even though no HTTP request was ever made.

Expected Behavior: A Markdown parse error triggers a fallback retry (e.g. the same content without `parse_mode`) rather than a silent drop; an empty message is reported as not sent.

Actual Behavior: A parse error is treated identically to any other terminal failure, and an empty message is reported as a successful send.

Suggested fix: Return `False` immediately for empty content with no event published; on a >=300 response whose body indicates a Markdown parse error, retry the same chunk once without `parse_mode` before giving up, leaving any other 4xx/5xx terminal as before. A local patch doing both, plus regression tests, is available if useful.

havolline1:
## Summary
- Fixes #783 — two bugs in `TelegramChannel.send()`:
  1. **Empty content silently reports success.** `textwrap.wrap("")` returns an empty list, so the send loop never made any HTTP request, yet the method still fell through to `self._publish_sent(...)` and `return True` — reporting a message as sent when nothing was ever transmitted.
  2. **Markdown parse errors drop messages outright.** Telegram rejects unparseable Markdown (lone asterisks, unclosed code fences, unescaped snake_case identifiers, etc.) with an HTTP 400 whose body names the cause (`"can't parse entities: ..."`). The old code treated this identically to every other 4xx/5xx and dropped the message with no fallback.
- Fix: return `False` immediately when there are no chunks to send. On a `>=300` response, check specifically for `"can't parse entities"` in the body — only for that failure mode, retry the same chunk once as plain text (parse_mode omitted) before giving up. Any other 4xx/5xx still fails immediately and is not retried, so a real problem (bad token, rate limit, etc.) isn't masked.

## Test plan
- [x] `test_send_empty_content_returns_false_without_request`: empty content must return `False`, make no HTTP call, and publish no `CHANNEL_MESSAGE_SENT` event.
- [x] `test_send_retries_without_markdown_on_parse_error`: a 400 with a `"can't parse entities"` body triggers exactly one retry without `parse_mode`, and a subsequent 200 succeeds.
- [x] `test_send_non_markdown_failure_does_not_retry`: a 500 (or any non-Markdown failure) must still fail immediately with no retry, so this fix doesn't mask unrelated errors.
- [x] Confirmed all three fail against the old code (false `True`, false `False`, respectively for the first two — the third already passed as a useful negative control) and pass after the fix.
- [x] `uv run pytest tests/channels/test_telegram.py` — 27 passed
- [x] `uv run pytest tests/channels/` — 782 passed, 1 pre-existing unrelated failure (`whatsapp_baileys` trying to spawn a Node.js subprocess not installed in this environment — nothing to do with Telegram)

### DIFF
--- a/src/openjarvis/channels/telegram.py
+++ b/src/openjarvis/channels/telegram.py
@@ -127,6 +127,12 @@ class TelegramChannel(BaseChannel):
                 break_long_words=True,
                 replace_whitespace=False,
             )
+            if not chunks:
+                # Empty content wraps to an empty chunk list -- there is
+                # nothing to send, so this must not fall through to the
+                # success path below and report a message that was never
+                # transmitted (#783).
+                return False
             for chunk in chunks:
                 payload: Dict[str, Any] = {
                     "chat_id": chat_id,
@@ -139,12 +145,28 @@ class TelegramChannel(BaseChannel):
 
                 resp = httpx.post(url, json=payload, timeout=10.0)
                 if resp.status_code >= 300:
-                    logger.warning(
-                        "Telegram API returned status %d: %s",
-                        resp.status_code,
-                        resp.text,
-                    )
-                    return False
+                    # Telegram rejects unparseable Markdown (lone asterisks,
+                    # unclosed code fences, unescaped snake_case
+                    # identifiers, etc.) with a 400 naming the cause in the
+                    # response body. Retry once as plain text instead of
+                    # dropping the message outright (#783).
+                    if self._parse_mode and "can't parse entities" in resp.text.lower():
+                        logger.warning(
+                            "Telegram rejected Markdown formatting, "
+                            "retrying as plain text: %s",
+                            resp.text,
+                        )
+                        plain_payload = {
+                            k: v for k, v in payload.items() if k != "parse_mode"
+                        }
+                        resp = httpx.post(url, json=plain_payload, timeout=10.0)
+                    if resp.status_code >= 300:
+                        logger.warning(
+                            "Telegram API returned status %d: %s",
+                            resp.status_code,
+                            resp.text,
+                        )
+                        return False
             self._publish_sent(channel, content, conversation_id)
             return True
         except Exception:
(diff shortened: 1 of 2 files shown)


ElliotSlusky (approved):
Maintainer review at the current head: approved. Telegram channel suite: 27 passed. Empty sends no longer report success and parse failures receive one safe plaintext retry. Repository rules still require a code-owner approval before merge.

PR #806, opened 2026-08-18
fix(connectors): don't advance last_sync on failed/in-progress checkpoints

### LINKED ISSUE #782: A failed connector sync advances last_sync, permanently skipping items
Description: `SyncEngine.sync` derives the incremental fetch window from `checkpoint["last_sync"]`, and `_save_checkpoint` unconditionally stamps `last_sync = datetime.now(utc)` on every call (src/openjarvis/connectors/sync_engine.py:97-175) - including on the failure path, which calls `_save_checkpoint(..., error=str(exc))` right before re-raising. A sync that dies partway through (e.g. at document 5,000 of 50,000) therefore records `last_sync` as the time of the failure; the next run asks the connector for items after that instant, so every item the failed run never got to is skipped forever, with no indication in the UI that anything was missed.

```python
except Exception as exc:
    self._save_checkpoint(
        connector_id,
        prior_items + items_ingested,
        cursor=current_cursor,
        error=str(exc),
    )
    raise
```
```python
def _save_checkpoint(self, connector_id, items_synced, *, cursor=None, error=None):
    now = datetime.now(tz=timezone.utc).isoformat()
    self._conn.execute(
        "INSERT INTO sync_state (connector_id, items_synced, cursor, last_sync, error) VALUES (?, ?, ?, ?, ?) ...",
        ...
    )
```

Steps to Reproduce:
1. Configure a connector whose `sync()` generator yields some documents, then raises an exception partway through.
2. Run `SyncEngine.sync(connector)` and let it fail; record the resulting `last_sync` from `get_checkpoint`.
3. Fix the underlying failure and run `SyncEngine.sync(connector)` again against the same connector, with a mock/fixture that would return the earlier, never-fetched items if asked for anything before the recorded `last_sync`.
4. Observe the retry starts from the failure-time watermark, never revisiting the items the failed run didn't reach.

Expected Behavior: A failed sync's checkpoint preserves the previous, successful watermark (or leaves it unset on a first-ever failed sync), so a retry re-covers the window the failed run didn't complete.

Actual Behavior: Every checkpoint write - including the failure path - stamps `last_sync` to the current time, permanently losing anything after the point of failure.

Suggested fix: Capture the sync's start time before the first fetch, and only advance `last_sync` on the final, successful checkpoint (using an UPSERT that leaves `last_sync` untouched on batch/error checkpoints, e.g. `last_sync = COALESCE(excluded.last_sync, sync_state.last_sync)`), using the start time rather than the completion time so items created during a long sync aren't skipped either. A local patch doing this, plus regression tests, is available if useful.

havolline1:
## Summary
- Fixes #782: `SyncEngine._save_checkpoint` unconditionally set `last_sync = now()` on *every* call — including mid-batch progress saves (every 100 items) and the error path when a sync raises. When a sync died partway through (e.g. after 5,000 of 50,000 documents), the failure time got recorded as `last_sync`. The next retry then called `connector.sync(since=<failure time>, ...)`, permanently skipping everything the failed attempt never reached — with no warning anywhere.
- `_save_checkpoint` now takes an optional `last_sync` parameter that's only passed on a successful sync completion; the UPSERT uses `last_sync = COALESCE(excluded.last_sync, sync_state.last_sync)` so progress and error checkpoints preserve the existing watermark instead of overwriting it with "now".
- The successful-completion watermark is the sync's **start** time (captured before the first fetch), not its end time, so documents created/modified while a long sync was running are still covered by the *next* sync's `since` filter rather than silently missed.

## Test plan
- [x] Added `test_failed_sync_does_not_advance_last_sync`: a connector that raises partway through must leave `last_sync` as `None` (this connector's first-ever sync) and record the error.
- [x] Added `test_retry_after_failure_does_not_skip_the_failed_window`: end-to-end — after a failed sync, a retry must see the *same* `since` as the original attempt (`None`), and only advances `last_sync` once the retry actually succeeds.
- [x] Confirmed both fail against the old code (`last_sync` was a fresh timestamp instead of `None`, and the retry received a real datetime instead of `None`) and pass after the fix.
- [x] `uv run pytest tests/connectors/test_sync_engine.py` — 6 passed
- [x] `uv run pytest tests/connectors/` — 354 passed, 5 pre-existing unrelated failures (`*_live*` tests requiring real API credentials for Oura/Strava/Spotify/Google Tasks/Obsidian — empty bearer tokens, nothing to do with this change)

### DIFF
--- a/src/openjarvis/connectors/sync_engine.py
+++ b/src/openjarvis/connectors/sync_engine.py
@@ -100,6 +100,11 @@ class SyncEngine:
             except (ValueError, TypeError):
                 pass
 
+        # Captured before the first fetch, not at completion, so items
+        # created/modified while this sync was running are still covered
+        # by the *next* sync's `since` filter instead of being skipped.
+        sync_started_at = datetime.now(tz=timezone.utc).isoformat()
+
         items_ingested = 0
         current_cursor: Optional[str] = prior_cursor
 
@@ -113,6 +118,12 @@ class SyncEngine:
                 if len(batch) >= _BATCH_SIZE:
                     items_ingested += self._pipeline.ingest(batch)
                     batch = []
+                    # Progress checkpoint: track cursor/items so a later
+                    # retry can resume from here, but must NOT advance
+                    # last_sync -- this sync hasn't completed yet, and
+                    # advancing the watermark on a checkpoint that isn't a
+                    # successful completion would permanently skip
+                    # whatever wasn't reached if the sync then fails (#782).
                     self._save_checkpoint(
                         connector_id,
                         prior_items + items_ingested,
@@ -132,12 +143,14 @@ class SyncEngine:
             )
             raise
 
-        # Final checkpoint — clear any previous error.
+        # Final checkpoint on successful completion — clear any previous
+        # error and advance the watermark to when this sync started.
         self._save_checkpoint(
             connector_id,
             prior_items + items_ingested,
             cursor=current_cursor,
             error=None,
+            last_sync=sync_started_at,
         )
         return items_ingested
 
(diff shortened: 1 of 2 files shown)


ElliotSlusky (approved):
Maintainer review at the current head: approved. Sync-engine suite: 6 passed. Failed/progress checkpoints preserve the old watermark and only a successful sync advances it. Repository rules still require a code-owner approval before merge.

PR #807, opened 2026-08-18
fix(tools): extend shell_exec env allowlist with Windows essentials

### LINKED ISSUE #789: shell_exec's sanitised env breaks DNS and provisions a Python install on Windows
Description: `shell_exec` builds a sanitised environment from a POSIX-only allow-list (`src/openjarvis/tools/shell_exec.py:24`):

```python
_BASE_ENV_KEYS = ("PATH", "HOME", "USER", "LANG", "TERM")
```

On Windows this drops everything cmd.exe and the CRT need. Two consequences are worth separating:

1. Without `SystemRoot`, any command that resolves DNS (`git clone/fetch`, `curl`, `pip install`, `npm install`) fails with `getaddrinfo() thread failed to start` even though the command itself is fine, because Winsock cannot initialise.
2. Without `LOCALAPPDATA`, launching `python` through the tool makes the Windows Store launcher treat the process as having no user install root and **provision a whole interpreter into the current working directory** - a `Python\` folder of roughly 170 MB appears inside the user's project. `PATHEXT` is also required for cmd.exe to resolve `git` / `npm` to their `.exe` / `.cmd`.

This is currently masked on a default install because `shell_exec` short-circuits to the Rust fast path (`_rust.ShellExecTool().execute(command, working_dir)`), whose PyO3 signature carries neither the environment nor the timeout and which therefore runs with the full inherited environment. The sanitised path becomes live as soon as the fast path is unavailable or bypassed.

Steps to Reproduce (Windows):
1. Force the Python path (uninstall/skip the Rust extension, or call the sanitised branch directly).
2. `ShellExecTool().execute(command='git ls-remote https://github.com/open-jarvis/OpenJarvis')` - fails with `getaddrinfo() thread failed to start`.
3. `ShellExecTool().execute(command='python -c "print(1)"')` from a project directory - observe a `Python\` directory (~170 MB) created in the working directory.

Expected Behavior: the sanitised environment still contains the OS baseline a Windows process needs, so commands behave normally and nothing is provisioned into the working directory.

Actual Behavior: DNS-resolving commands fail, and running `python` writes a full interpreter into the current directory.

Environment: OpenJarvis @ 1ca3ec6 (main), Python 3.13.1, Windows 11. Windows-only.

Suggested fix: Extend the allow-list on Windows with `SystemRoot`, `SystemDrive`, `COMSPEC`, `PATHEXT`, `TEMP`, `TMP`, `USERPROFILE`, `LOCALAPPDATA` and `APPDATA`. A local patch with a regression test (asserting the passthrough, and that running python leaves no interpreter behind in the working directory) is available if useful.

havolline1:
## Summary
- Fixes #789: `shell_exec`'s Python-fallback env allowlist (`_BASE_ENV_KEYS`, used when the Rust fast path is unavailable) is POSIX-focused — `"PATH", "HOME", "USER", "LANG", "TERM"` — and strips variables Windows needs for basic process bootstrapping:
  - Missing `SystemRoot` breaks Winsock/DNS initialization, so any network-touching command (`git`, `curl`, `pip`, `npm`) fails with `"getaddrinfo() thread failed to start"`.
  - Missing `LOCALAPPDATA` makes the Windows Store Python launcher interpret the process as having no user install root, provisioning a fresh ~170MB `Python\` folder in the cwd instead of finding the real interpreter.
- Added `SystemRoot`, `SystemDrive`, `COMSPEC`, `PATHEXT`, `TEMP`, `TMP`, `USERPROFILE`, `LOCALAPPDATA`, and `APPDATA` to the allowlist, matching the issue's suggested fix exactly. These are harmless on POSIX — `os.environ.get()` returns `None` for unset keys and they're skipped, same as before.

## Test plan
- [x] Added `test_base_env_keys_include_windows_essentials`: asserts all nine Windows keys are present in `_BASE_ENV_KEYS`.
- [x] Added `test_windows_env_vars_reach_the_subprocess`: forces the Rust backend unavailable (`ImportError`) to exercise the Python fallback, sets `SystemRoot`/`LOCALAPPDATA` via `monkeypatch.setenv`, and asserts they actually reach `subprocess.run`'s `env=` kwarg.
- [x] Confirmed both fail against the old code (this dev environment is genuinely Windows, so the failure output shows the real host `PATH`/`HOME` with no `SystemRoot` present) and pass after the fix.
- [x] `uv run pytest tests/tools/test_shell_exec.py` — 19 passed, 4 pre-existing skips (unrelated: Rust-backend timeout/env-isolation limitations, not this fix)
- [x] `uv run pytest tests/tools/` — 668 passed

### DIFF
--- a/src/openjarvis/tools/shell_exec.py
+++ b/src/openjarvis/tools/shell_exec.py
@@ -20,8 +20,30 @@ _MAX_TIMEOUT = 300
 # Default timeout (seconds)
 _DEFAULT_TIMEOUT = 30
 
-# Environment variables always passed through
-_BASE_ENV_KEYS = ("PATH", "HOME", "USER", "LANG", "TERM")
+# Environment variables always passed through. The Windows-specific
+# entries are harmless on POSIX (os.environ.get() returns None for them
+# and they're skipped below) but essential on Windows: missing SystemRoot
+# breaks Winsock/DNS initialization for any network-touching command
+# (git, curl, pip, npm all fail with "getaddrinfo() thread failed to
+# start"), and missing LOCALAPPDATA makes the Windows Store Python
+# launcher provision a fresh ~170MB Python\ folder in the cwd instead of
+# finding the real interpreter (#789).
+_BASE_ENV_KEYS = (
+    "PATH",
+    "HOME",
+    "USER",
+    "LANG",
+    "TERM",
+    "SystemRoot",
+    "SystemDrive",
+    "COMSPEC",
+    "PATHEXT",
+    "TEMP",
+    "TMP",
+    "USERPROFILE",
+    "LOCALAPPDATA",
+    "APPDATA",
+)
 
 
 @ToolRegistry.register("shell_exec")
--- a/src/openjarvis/tools/shell_exec.py
+++ b/src/openjarvis/tools/shell_exec.py
@@ -20,20 +20,21 @@ _MAX_TIMEOUT = 300
 # Default timeout (seconds)
 _DEFAULT_TIMEOUT = 30
 
-# Environment variables always passed through. The Windows-specific
-# entries are harmless on POSIX (os.environ.get() returns None for them
-# and they're skipped below) but essential on Windows: missing SystemRoot
-# breaks Winsock/DNS initialization for any network-touching command
-# (git, curl, pip, npm all fail with "getaddrinfo() thread failed to
-# start"), and missing LOCALAPPDATA makes the Windows Store Python
-# launcher provision a fresh ~170MB Python\ folder in the cwd instead of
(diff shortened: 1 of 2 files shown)


ElliotSlusky:
Pushed follow-up commit b7d85ed. Windows-only environment keys are now conditionally included only on Windows. Added a real Windows child-process probe that launches the active interpreter through the sanitized fallback, verifies SystemRoot and LOCALAPPDATA propagation, and resolves localhost through Winsock. Local verification: shell-exec suite 19 passed, 6 Windows-only skips; Ruff check and format check passed.

ElliotSlusky (approved):

PR #808, opened 2026-08-18
fix(channels): make disconnect() actually stop the Telegram/Discord listener

### LINKED ISSUE #784: Channel disconnect() lies about thread status, allowing duplicate pollers
Description: `TelegramChannel.disconnect()` sets `_stop_event` and joins the listener thread with a 5-second timeout, but `_poll_loop` never observes that event - `app.run_polling()` blocks until its own event loop stops, so the join always times out. `disconnect()` still sets `_listener_thread = None` and `_status = DISCONNECTED` regardless, while the thread keeps polling in the background (src/openjarvis/channels/telegram.py:87-93, 216). `DiscordChannel` has the identical defect (`loop.run_until_complete(client.start(...))`, discord_channel.py:75-80, 191-193) and additionally never closes the event loop it created. A subsequent `connect()` then starts a second poller: Telegram answers `409 Conflict: terminated by other getUpdates request` to both pollers, and Discord opens a second gateway connection and handles every incoming message twice.

```python
def disconnect(self) -> None:
    """Stop the listener thread."""
    self._stop_event.set()
    if self._listener_thread is not None:
        self._listener_thread.join(timeout=5.0)
        self._listener_thread = None
    self._status = ChannelStatus.DISCONNECTED
```
```python
app.add_handler(MessageHandler(filters.TEXT, _handle_msg))
app.run_polling(stop_signals=None, drop_pending_updates=True)  # blocks; never checks _stop_event
```

Steps to Reproduce:
1. `channel = TelegramChannel(bot_token="...")`; `channel.connect()` (starts the long-polling thread).
2. `channel.disconnect()` and check `channel.status()` immediately afterward - observe it reports `DISCONNECTED` in under 5 seconds, but the listener thread is actually still alive and still polling (verify with `threading.enumerate()` or the underlying `Application`'s state).
3. Call `channel.connect()` again.
4. Observe two independent pollers running concurrently against the same bot token, producing `409 Conflict` responses from Telegram (or duplicate message handling for Discord).

Expected Behavior: `disconnect()` actually stops the listener before returning, or at minimum does not report `DISCONNECTED` while the thread is still alive; `connect()` refuses to start a second listener while one is already running.

Actual Behavior: `disconnect()` always reports success after the join timeout regardless of whether the thread stopped, and nothing prevents starting a second listener on top of a still-running one.

Suggested fix: Have the listener thread publish its `Application`/`Client` and its own event loop on the instance, and have `disconnect()` actually ask the listener to stop from inside its own loop (`loop.call_soon_threadsafe(app.stop_running)` for Telegram, `asyncio.run_coroutine_threadsafe(client.close(), loop)` for Discord, with Discord also closing the loop) before joining - only reporting `DISCONNECTED` if the thread actually died. Have `connect()` refuse to start a second listener while one is alive. A local patch doing this for both channels, plus regression tests, is available if useful.

havolline1:
## Summary
- Fixes #784: both `TelegramChannel` and `DiscordChannel` set `_stop_event` in `disconnect()`, but neither listener loop ever observed it — `Application.run_polling()` and `client.start()` block on their own event loop regardless of that event. The 5-second thread join always timed out, and both channels set status to `DISCONNECTED` unconditionally afterward, regardless of whether the thread had actually terminated. A subsequent `connect()` then spawned a *second* listener against the same bot token: Telegram's `getUpdates` API returns 409 Conflict, and Discord duplicates message handling.
- **Telegram**: `_poll_loop` now creates and publishes its own event loop *before* building the `Application` (so `run_polling()`'s internal `asyncio.get_event_loop()` picks up the already-set loop instead of creating its own). `disconnect()` schedules `Application.stop_running()` onto that loop via `loop.call_soon_threadsafe()`. I verified via the actual installed `python-telegram-bot` source that `stop_running()` calls `asyncio.get_running_loop().stop()` internally — that's only safe from *within* the loop's own thread, hence the threadsafe scheduling rather than a direct call.
- **Discord**: `_gateway_loop` already created its own event loop; `disconnect()` now schedules `client.close()` (a coroutine, confirmed via the installed `discord.py`) onto it via `asyncio.run_coroutine_threadsafe()` and waits for it to actually finish (`future.result(timeout=5.0)`).
- Both `disconnect()` implementations now only report `DISCONNECTED` if the listener thread actually terminated within the join timeout — if not, status is left unchanged (so a caller can tell the stop didn't succeed, rather than being lied to).
- Both `connect()` implementations now guard against starting a duplicate listener while one is already running and alive.

## Test plan
- [x] `TestDisconnectStopsRealListener` (both `test_telegram.py` and `test_discord_channel.py`, 4 tests each): stop mechanism is scheduled correctly (`call_soon_threadsafe`/`run_coroutine_threadsafe`) with the right args; status stays non-`DISCONNECTED` when the thread is still alive after the join; the old join-only behavior still works when `_app`/`_client`/`_loop` are `None` (send-only mode, library not installed); `connect()` refuses to start a second thread while one is already alive.
- [x] Confirmed all the new assertions fail against the old code (status incorrectly flips to `DISCONNECTED` regardless of thread state; `connect()` spawns a duplicate thread) and pass after the fix.
- [x] `uv run pytest tests/channels/test_telegram.py tests/channels/test_discord_channel.py` — 51 passed
- [x] `uv run pytest tests/channels/` — 790 passed, 1 pre-existing unrelated failure (`whatsapp_baileys` — missing Node.js subprocess in this environment)
- [x] Ran the full suite with `python-telegram-bot`/`discord.py` actually installed (`uv sync --extra dev --extra framework-comparison --extra server --extra channel-telegram --extra channel-discord`) to make sure adding real imports to these paths doesn't break anything elsewhere — 7375 passed; the 161 failures are the same pre-existing environment issues verified unrelated in earlier PRs on this repo (missing compiled `openjarvis_rust` extension, Linux-only RAPL sysfs paths, FTS5, etc.) — none touch `channels/`.

Note: `python-telegram-bot`/`discord.py` aren't installed by CI's main test job (only `--extra dev --extra framework-comparison --extra server`), so `connect()` falls into the "library not installed, send-only mode" branch there — the new tests exercise the real fix logic directly via mocking rather than depending on those extras being present.


ElliotSlusky:
Pushed dcb7aa8 to make channel shutdown race-safe. Discord now closes its client and event loop, Telegram closes its loop, and both abort startup if disconnect wins before resource publication. Added lifecycle tests for disconnect-before-publication. Validation: 53 channel tests passed; Ruff passed.

ElliotSlusky (dismissed):

PR #809, opened 2026-08-18
fix(skills): prune cyclic skills instead of crashing discovery

### LINKED ISSUE #781: A dependency cycle between two skills crashes skill discovery entirely
Description: `SkillManager.discover()` calls `validate_dependencies(self._skills)` unguarded (src/openjarvis/skills/manager.py:87-89). A dependency cycle between even one pair of installed skills raises `DependencyCycleError` uncaught, which crashes `jarvis skill list`/`info`/`run`/`discover` in the CLI with a raw traceback (the CLI has no try/except around it), and in `SystemBuilder` the whole skills block sits inside one broad `except`, silently dropping every skill (not just the cyclic pair) and leaving `skill_manager = None`.

```python
if paths:
    for directory in paths:
        manifests = discover_skills(directory)
        for manifest in manifests:
            if manifest.name not in self._skills:
                self._skills[manifest.name] = manifest

    # Validate the dependency graph after loading skills
    if self._skills:
        validate_dependencies(self._skills)
```

Steps to Reproduce:
1. Install two skills, A and B, where A declares a dependency on B and B declares a dependency on A (a two-node cycle).
2. Call `SkillManager().discover([skills_dir])` (or run `jarvis skill list`).
3. Observe `DependencyCycleError` propagates uncaught, crashing the command - with every other, non-cyclic skill also lost when this happens inside `SystemBuilder`'s broad `except`.

Expected Behavior: A dependency cycle between a pair of skills drops just that pair (with a logged warning) and leaves every other, acyclic skill registered and usable.

Actual Behavior: The whole discovery call crashes, and depending on the caller, every skill (not just the cyclic ones) can be lost.

Suggested fix: Detect skills with no valid topological position (e.g. via a Kahn's-algorithm pass over the dependency graph), log and drop just those, repeating in case dropping one cycle exposes another independent cycle, then continue with the remaining acyclic skills. A local patch doing this, plus regression tests, is available if useful.

havolline1:
## Summary
- Fixes #781: `SkillManager.discover()` called `validate_dependencies(self._skills)` unguarded, so a dependency cycle between even one pair of skills raised `DependencyCycleError` uncaught. This propagated through both CLI commands and `SystemBuilder`, where a broad exception handler caught it and silently discarded **every** discovered skill, not just the cyclic pair.
- `DependencyCycleError` now carries a `skills: Set[str]` attribute — the set Kahn's algorithm identified as unresolved (`set(graph) - set(order)`) — instead of only encoding it into the message string, so callers can act on it programmatically.
- New `SkillManager._prune_dependency_cycles()` iteratively removes exactly the skills each cycle involves and re-validates. Repeating (rather than a single pass) matters because removing a cyclic skill can resolve others that only looked stuck due to an edge pointing into the cycle — `build_dependency_graph` already silently drops edges to skills no longer present in the dict, so once the cyclic skills are gone, dependents on them just lose that edge instead of failing again. Acyclic skills remain discoverable and usable; cyclic ones are dropped with a `logger.warning`.
- No changes needed in `SystemBuilder` — it just calls `SkillManager.discover()`, so fixing the root cause there fixes both the CLI and builder paths the issue mentions.

## Test plan
- [x] `test_cycle_error_carries_cyclic_skill_names` (`tests/skills/test_dependency.py`): asserts `DependencyCycleError.skills` matches the actual cyclic pair.
- [x] `TestSkillManagerDependencyCycles` (`tests/skills/test_manager.py`, 3 tests): a 2-skill cycle no longer raises out of `discover()`; an unrelated acyclic skill discovered alongside a cyclic pair survives and remains fully resolvable; a 3-way cycle (a→c→b→a) is pruned entirely, leaving only the unrelated skill.
- [x] Confirmed all four fail against the old code (`AttributeError: no attribute 'skills'`, and uncaught `DependencyCycleError` crashing `discover()`) and pass after the fix.
- [x] `uv run pytest tests/skills/test_dependency.py tests/skills/test_manager.py` — 54 passed
- [x] `uv run pytest tests/skills/` — 281 passed, 4 pre-existing unrelated failures (`test_integration_live.py` — requires a real running inference engine, e.g. `ollama serve`; matches the `-m "not live"` category CI already excludes)

### DIFF
--- a/src/openjarvis/skills/dependency.py
+++ b/src/openjarvis/skills/dependency.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from collections import deque
-from typing import TYPE_CHECKING, Dict, List, Set
+from typing import TYPE_CHECKING, Dict, List, Optional, Set
 
 if TYPE_CHECKING:
     from openjarvis.skills.types import SkillManifest
@@ -11,6 +11,13 @@ if TYPE_CHECKING:
 
 class DependencyCycleError(Exception):
     """Raised when a cycle is detected in the skill dependency graph."""
+
+    def __init__(self, message: str, skills: Optional[Set[str]] = None) -> None:
+        super().__init__(message)
+        # The skills involved in (or downstream of) the cycle, so callers
+        # can prune exactly those rather than parsing the message string
+        # or discarding the whole graph (#781).
+        self.skills: Set[str] = skills if skills is not None else set()
 
 
 class DepthExceededError(Exception):
@@ -101,9 +108,11 @@ def validate_dependencies(
                 queue.append(dependent)
 
     if len(order) != len(graph):
+        cyclic = set(graph) - set(order)
         raise DependencyCycleError(
             "Cycle detected in skill dependency graph. "
-            f"Skills involved: {set(graph) - set(order)}"
+            f"Skills involved: {cyclic}",
+            skills=cyclic,
         )
 
     # --- Depth enforcement via DFS ---
--- a/src/openjarvis/skills/manager.py
+++ b/src/openjarvis/skills/manager.py
@@ -3,17 +3,20 @@
 from __future__ import annotations
 
 import html
+import logging
 from pathlib import Path
 from typing import Any, Dict, List, Optional
 
 from openjarvis.core.events import EventBus
 from openjarvis.core.paths import get_config_dir
-from openjarvis.skills.dependency import validate_dependencies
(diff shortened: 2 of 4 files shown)


ElliotSlusky (approved):
Maintainer review at the current head: approved. Dependency and manager suites: 54 passed. Cyclic/unusable dependency sets are pruned while independent skills remain available. Repository rules still require a code-owner approval before merge.

PR #810, opened 2026-08-18
fix(skills): log discover_skills() failures instead of swallowing them

### LINKED ISSUE #780: Malformed skills silently vanish from jarvis skill list, unlogged
Description: All three discovery branches of `discover_skills` (flat `*.toml` files, direct skill packages, sourced `<source>/<name>/` packages) use a bare `except Exception: continue` with no logging at any level (src/openjarvis/skills/loader.py:280-309). Any skill with a malformed `skill.toml`/`SKILL.md`, an invalid name, or a missing required field silently disappears from `jarvis skill list`/`jarvis skill info`/the model's tool catalog with zero diagnostics - inconsistent with other loaders in the same codebase (`_read_source_metadata`, `SkillOverlayLoader.load`) which both log a warning on the same class of failure.

```python
for toml_file in sorted(directory.glob("*.toml")):
    try:
        manifests.append(load_skill(toml_file))
    except Exception:
        continue
...
        try:
            manifests.append(load_skill_directory(child))
        except Exception:
            continue
```

Steps to Reproduce:
1. Create a skills directory containing one valid skill and one skill with a deliberately malformed `skill.toml` (e.g. invalid TOML syntax, or an invalid name with an uppercase letter).
2. Call `discover_skills(directory)`.
3. Observe the malformed skill is simply absent from the returned list, and nothing in the logs (at any level) indicates it was skipped or why.

Expected Behavior: A skill that fails to load is skipped but logged (path + exception) so a user can debug why it's missing from the catalog.

Actual Behavior: It vanishes silently with no diagnostic anywhere.

Suggested fix: Log a warning (path and exception) in each of the three `except Exception: continue` blocks before continuing, matching the logging style already used elsewhere in the skills loader. A local patch doing this, plus regression tests, is available if useful.

havolline1:
## Summary
- Fixes #780: all three discovery branches in `discover_skills()` (flat `*.toml` files, direct skill directories, sourced-layout directories) used a bare `except Exception: continue` with no logging. Any skill with a malformed `skill.toml`/`SKILL.md`, invalid TOML syntax, or a missing required field silently vanished from the returned list with zero diagnostic output at any log level — unlike `_read_source_metadata` and `SkillOverlayLoader.load` elsewhere in the same module, which both log a warning on similar failures.
- Added `LOGGER.warning("Failed to load skill from %s: %s", path, exc)` to each of the three handlers before `continue`, matching the module's existing pattern exactly.

## Test plan
- [x] Added `tests/skills/test_loader_discover.py` (`TestDiscoverSkillsLogsFailures`, 4 tests): a malformed flat `.toml` file, a malformed skill directory, and a malformed sourced-layout skill directory each produce a warning-level log record naming the failing path; a valid skill discovered alongside a malformed one still loads successfully while the malformed one's failure is still logged.
- [x] Confirmed all four fail against the old code (`assert False` — no log records at all) and pass after the fix.
- [x] `uv run pytest tests/skills/test_loader_discover.py` — 4 passed
- [x] `uv run pytest tests/skills/` — 285 passed, 4 pre-existing unrelated failures (`test_integration_live.py` — requires a real running inference engine, e.g. `ollama serve`)

Branched directly off `main`; applies cleanly independent of #809 (the dependency-cycle pruning fix for #781, also currently open) — that PR touches `dependency.py`/`manager.py`, this one only touches `loader.py`.

### DIFF
--- a/src/openjarvis/skills/loader.py
+++ b/src/openjarvis/skills/loader.py
@@ -281,7 +281,8 @@ def discover_skills(directory: str | Path) -> list[SkillManifest]:
     for toml_file in sorted(directory.glob("*.toml")):
         try:
             manifests.append(load_skill(toml_file))
-        except Exception:
+        except Exception as exc:
+            LOGGER.warning("Failed to load skill from %s: %s", toml_file, exc)
             continue
 
     # Walk one or two levels deep looking for skill packages
@@ -292,7 +293,8 @@ def discover_skills(directory: str | Path) -> list[SkillManifest]:
         if (child / "skill.toml").exists() or (child / "SKILL.md").exists():
             try:
                 manifests.append(load_skill_directory(child))
-            except Exception:
+            except Exception as exc:
+                LOGGER.warning("Failed to load skill from %s: %s", child, exc)
                 continue
             continue
 
@@ -305,7 +307,10 @@ def discover_skills(directory: str | Path) -> list[SkillManifest]:
             ).exists():
                 try:
                     manifests.append(load_skill_directory(grandchild))
-                except Exception:
+                except Exception as exc:
+                    LOGGER.warning(
+                        "Failed to load skill from %s: %s", grandchild, exc
+                    )
                     continue
 
     return manifests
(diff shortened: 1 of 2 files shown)


ElliotSlusky (approved):
Maintainer review at the current head: approved. Focused discovery suite: 4 passed. The change adds diagnostics to all three silent failure paths without changing loading behavior. Repository rules still require a code-owner approval before merge.

PR #811, opened 2026-08-18
test(core): stop credential tests leaking TAVILY_API_KEY across the session

### LINKED ISSUE #788: Credential tests leak env vars, making data-boundary tests fail only in a full run
Description: `tests/core/test_credentials.py` exercises `save_credential()` and `inject_credentials()`, both of which set `os.environ[key]` by design (`src/openjarvis/core/credentials.py:170` and `:206`). The module never restores the process environment afterwards, so `TAVILY_API_KEY` stays set for every test that runs later in the same session. That is enough to make `_has_cloud_api_surface()` (`src/openjarvis/security/data_boundary_audit.py:1480-1483`, which returns `True` as soon as any `API_KEY_ENV_VARS` entry is set) report a cloud surface where the test expects none.

```python
def test_inject_credentials_restores_saved_value(cred_path, monkeypatch):
    save_credential("web_search", "TAVILY_API_KEY", "tvly-persisted", path=cred_path)
    monkeypatch.delenv("TAVILY_API_KEY")
    inject_credentials(path=cred_path)
    assert os.environ["TAVILY_API_KEY"] == "tvly-persisted"
```

Note the `monkeypatch.delenv` here records the value `save_credential` just wrote, so monkeypatch's undo *re-sets* it at teardown even if the module adds its own cleanup fixture (monkeypatch's finalizer runs after an autouse fixture declared in the module).

Steps to Reproduce:
1. `pytest tests/core/test_credentials.py tests/security/test_data_boundary_audit.py -q`
2. Observe `test_local_engine_vendor_model_names_are_not_cloud[deepseek-r1:7b]` (and the two other parametrisations) fail with `assert 'frontend-credential-storage-not-inspected' not in {...}`.
3. Run either file alone and both pass.

Expected Behavior: test modules do not mutate the process environment for the rest of the session; the data-boundary tests give the same result whichever other tests ran first.

Actual Behavior: `TAVILY_API_KEY` (and `EMAIL_PASSWORD`, if credential tests for the email section are added) leak, and three data-boundary assertions fail only in a full-suite run.

Suggested fix: Add an autouse fixture to the module that snapshots `dict(os.environ)` and restores it on teardown, and drop the environment variable directly (`os.environ.pop(...)`) rather than through `monkeypatch.delenv` in the one test that removes a value `save_credential` has just written. A local patch plus the ordering note above is available if useful.

havolline1:
## Summary
- Fixes #788: `save_credential()`/`inject_credentials()`/`delete_credential()` mutate `os.environ` directly as a side effect of file persistence — they're production code, not test doubles, so `monkeypatch`'s undo stack has no idea those mutations happened. A test using `monkeypatch.delenv("TAVILY_API_KEY")` to clear the var before calling one of these functions would have `monkeypatch` snapshot whatever value had *already leaked in from an earlier test* and then **restore that leaked value at teardown** — permanently re-leaking it into every later test in the session, including `tests/security/test_data_boundary_audit.py`'s cloud-surface detection, which treats any set API-key env var as a cloud API surface present.
- Verified the exact mechanism directly: running just `test_save_and_load` (which calls `save_credential` with no `monkeypatch` involvement at all) leaves `TAVILY_API_KEY=tvly-123` in `os.environ` for the rest of the pytest session — confirmed via a throwaway probe test run right after it.
- Added an autouse `_isolate_environ` fixture that snapshots the full `os.environ` before each test in this module and force-restores it after, regardless of what `save_credential`/`monkeypatch` did in between. Replaced the two `monkeypatch.delenv()` calls with direct `os.environ.pop(..., None)`, since `monkeypatch`'s own restore semantics are now redundant (and were the actual mechanism re-leaking the value) once the module-level fixture owns full restoration.

## Test plan
- [x] Added `test_zz_previous_test_did_not_leak_env`, which runs immediately after `test_save_and_load` (pytest's default file-order execution) and asserts `TAVILY_API_KEY` is absent from `os.environ`.
- [x] Confirmed it fails against the old code with the exact leaked value (`'TAVILY_API_KEY': 'tvly-123'` visible in the failure's `os.environ` dump) when the new fixture's restore logic is temporarily disabled, and passes with the fix in place.
- [x] Re-ran the original minimal repro (`test_save_and_load` + an external probe test) directly — confirmed `TAVILY_API_KEY leaked: None` after the fix, vs. `TAVILY_API_KEY leaked: tvly-123` before.
- [x] `uv run pytest tests/core/test_credentials.py -k "not test_file_permissions"` — 10 passed (`test_file_permissions` is a separate, pre-existing, unrelated Windows POSIX-chmod-on-NTFS failure — not caused by or related to this fix)
- [x] `uv run pytest tests/core/test_credentials.py tests/security/test_data_boundary_audit.py -k "not test_file_permissions"` — 170 passed
- [x] `uv run pytest tests/core/ -k "not test_file_permissions"` — 234 passed, 4 pre-existing unrelated failures (missing compiled `openjarvis_rust` extension, not built in this environment)
- [x] `uv run ruff check` on the changed file — clean

### DIFF
--- a/tests/core/test_credentials.py
+++ b/tests/core/test_credentials.py
@@ -5,6 +5,7 @@ import os
 import pytest
 
 from openjarvis.core.credentials import (
+    TOOL_CREDENTIALS,
     delete_credential,
     get_credential_status,
     inject_credentials,
@@ -12,15 +13,31 @@ from openjarvis.core.credentials import (
     save_credential,
 )
 
+_CREDENTIAL_ENV_KEYS = frozenset(
+    key for keys in TOOL_CREDENTIALS.values() for key in keys
+)
+_ENV_BASELINE = {
+    key: value for key, value in os.environ.items() if key in _CREDENTIAL_ENV_KEYS
+}
+
 
 @pytest.fixture
 def cred_path(tmp_path):
     return tmp_path / "credentials.toml"
 
 
+def _restore_credential_environ(snapshot):
+    """Restore credential variables without disturbing pytest/process internals."""
+    for key in _CREDENTIAL_ENV_KEYS:
+        if key in snapshot:
+            os.environ[key] = snapshot[key]
+        else:
+            os.environ.pop(key, None)
+
+
 @pytest.fixture(autouse=True)
 def _isolate_environ():
-    """Snapshot and restore os.environ around every test in this module.
+    """Reset os.environ to the collection baseline around every test.
 
     save_credential()/inject_credentials()/delete_credential() mutate
     os.environ directly as a side effect of file persistence -- they are
@@ -32,13 +49,13 @@ def _isolate_environ():
     leaked value at teardown -- permanently re-leaking it into every
     later test in the session, including tests/security/
     test_data_boundary_audit.py's cloud-surface detection (#788).
-    Force a full restore of the real environment after every test here
-    instead of relying on monkeypatch to track these direct mutations.
+    Resetting both before and after each test makes isolation independent
+    of test order while preserving legitimate ambient credentials supplied
+    by the developer or CI process.
     """
-    original = dict(os.environ)
+    _restore_credential_environ(_ENV_BASELINE)
     yield
-    os.environ.clear()
-    os.environ.update(original)
+    _restore_credential_environ(_ENV_BASELINE)
 
 
 def test_save_and_load(cred_path):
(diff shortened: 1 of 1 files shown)


ElliotSlusky:
Pushed aa23141 to make credential environment isolation baseline-aware and order-independent. The fixture now restores only supported credential variables, preserves legitimate ambient values such as TAVILY_API_KEY, and does not disturb pytest/process internals. Validation: 11 tests passed with no ambient key and 11 passed with TAVILY_API_KEY set; Ruff passed.

ElliotSlusky (approved):

PR #813, opened 2026-08-18
fix(tests): fix three Windows-only test failures

### LINKED ISSUE #785: Three tests fail on Windows: TOML escaping, POSIX paths, wiped env
Description: Three independent tests in the suite make Windows-incompatible assumptions, all failing when run on a Windows machine even though the code under test is correct. (1) `tests/evals/core/test_config_split_parsing.py:18` interpolates a raw Windows `tmp_path` (e.g. `C:\Users\...\out`) directly inside a TOML basic (double-quoted) string; TOML basic strings treat backslash as an escape introducer, so a segment like `\Users` parses as the invalid escape `\U`, and `tomllib` rejects it. (2) `tests/learning/test_trial_runner.py:140` asserts `cfg.output_path.startswith("out/")`, a POSIX-only literal, against a path built with `Path(self.output_dir) / f"..."`, which legitimately produces backslashes on Windows. (3) `tests/evals/comparison/test_openclaw_runner_contract.py:38-41` builds `env = {"OPENCLAW_PATH": ..., "HOME": ...}` and passes it as the entire environment to `subprocess.run(..., env=env)`, wiping `TEMP`/`TMP`/`SystemRoot`/`PATH`; Node's `os.tmpdir()` (used by the runner script under test) then falls back to the literal string `'undefined'` when none of those are set, producing an unusable temp path and an `ENOENT` from `mkdtempSync`.

```python
# test_config_split_parsing.py:18
output_dir = "{tmp_path / "out"}"
```
```python
# test_trial_runner.py:140
assert cfg.output_path.startswith("out/")
```
```python
# test_openclaw_runner_contract.py:36-41
env = {
    "OPENCLAW_PATH": str(tmp_path),
    "HOME": str(tmp_path / "home"),
}
result = subprocess.run([node, str(runner), ...], env=env, ...)
```

Steps to Reproduce (on Windows):
1. `pytest tests/evals/core/test_config_split_parsing.py` - observe all 5 tests fail with `tomllib.TOMLDecodeError: Invalid hex value (at line 6, column 19)`.
2. `pytest tests/learning/test_trial_runner.py -k test_output_path_contains_trial_id` - observe `AssertionError: assert False where False = 'out\\trial-abc_qwen3-8b.jsonl'.startswith('out/')`.
3. `pytest tests/evals/comparison/test_openclaw_runner_contract.py::test_openclaw_runner_parses_real_agent_json_shape` (requires `node` on PATH) - observe the assertion on `result.returncode == 0` fails with a Node `ENOENT`/`mkdtemp` stack trace.

Expected Behavior: The full test suite passes on Windows; none of these three assertions/fixtures should depend on POSIX-only path conventions or on the parent process's environment being irrelevant to the subprocess under test.

Actual Behavior: All three fail on Windows, for three distinct platform-specific reasons.

Environment: OpenJarvis @ 1ca3ec6 (main), Python 3.13.1, Windows 11, Node.js present on PATH. Windows-only - these tests pass as written on POSIX systems.

Suggested fix: (1) Render the TOML path with `.as_posix()` so it is unambiguous inside a TOML basic string on both platforms. (2) Replace the POSIX-literal `startswith("out/")` check with an `os.sep`-agnostic comparison, e.g. `Path(cfg.output_path).parts[0] == "out"`. (3) Build the child environment as `{**os.environ, "OPENCLAW_PATH": ..., "HOME": ..., "USERPROFILE": ...}` so the parent's `TEMP`/`TMP`/`SystemRoot`/`PATH` are inherited while still overriding the test-specific variables. A local patch making all three changes is available if useful.

havolline1:
## Summary
Fixes #785 — three independent platform assumptions that broke these tests on Windows while the underlying application code worked correctly:

1. **`test_config_split_parsing.py`**: a raw Windows path (`C:\Users\...\out`) was interpolated directly into a TOML basic string. TOML treats backslashes as escape introducers, so `\Users` parsed as the invalid escape `\U` and `tomllib` rejected the whole config with `TOMLDecodeError: Invalid hex value`. Fixed by rendering the path with `.as_posix()`.
2. **`test_trial_runner.py`**: asserted `cfg.output_path.startswith("out/")`, but `Path(...) / "..."` legitimately produces backslashes on Windows (`'out\trial-abc_qwen3-8b.jsonl'`). Fixed by comparing path components instead: `Path(cfg.output_path).parts[0] == "out"`.
3. **`test_openclaw_runner_contract.py`**: `subprocess.run`'s `env` kwarg **replaces** the environment rather than merging it, so passing only `{OPENCLAW_PATH, HOME}` wiped `TEMP`/`TMP`/`SystemRoot`/`PATH`. Node needs those to even initialize on Windows. Fixed by building env as `{**os.environ, ...overrides}` to inherit the parent's critical variables.

## Test plan
- [x] Confirmed all three fail on this Windows dev machine with the exact failures described in the issue (`tomllib.TOMLDecodeError: Invalid hex value`; `AssertionError` on `startswith("out/")`; for #3, reproduced as a Node native crash — `Assertion failed: ncrypto::CSPRNG(nullptr, 0)` — same root cause of a wiped Windows environment, different visible symptom depending on Node version, since this environment's Node 24 fails earlier during crypto init rather than at `mkdtempSync`).
- [x] Confirmed all three pass after their respective fixes.
- [x] `uv run pytest tests/evals/core/ tests/learning/test_trial_runner.py tests/evals/comparison/` — 103 passed, 3 skipped (unrelated skips)
- [x] `uv run ruff check` on all three changed files — clean

### DIFF
--- a/tests/evals/comparison/test_openclaw_runner_contract.py
+++ b/tests/evals/comparison/test_openclaw_runner_contract.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+import os
 import shutil
 import subprocess
 from pathlib import Path
@@ -35,9 +36,17 @@ def test_openclaw_runner_parses_real_agent_json_shape(tmp_path: Path) -> None:
         / "src/openjarvis/evals/backends/external/_runners/openclaw_runner.mjs"
     )
     out_json = tmp_path / "out.json"
+    # subprocess.run's `env` REPLACES the environment rather than merging
+    # it, so passing just these two keys wiped everything Node needs to
+    # even start on Windows (TEMP/TMP/SystemRoot/PATH) -- os.tmpdir()
+    # falls back to the literal string 'undefined' when those are unset,
+    # and mkdtemp then fails with ENOENT (#785). Inherit the parent
+    # environment and layer the test-specific overrides on top.
     env = {
+        **os.environ,
         "OPENCLAW_PATH": str(tmp_path),
         "HOME": str(tmp_path / "home"),
+        "USERPROFILE": str(tmp_path / "home"),
     }
     result = subprocess.run(
         [
--- a/tests/evals/core/test_config_split_parsing.py
+++ b/tests/evals/core/test_config_split_parsing.py
@@ -15,7 +15,7 @@ def _write_config(tmp_path: Path, split_value: str | None) -> Path:
 name = "split-parse-test"
 
 [run]
-output_dir = "{tmp_path / "out"}"
+output_dir = "{(tmp_path / "out").as_posix()}"
 seed = 42
 
 [[models]]
--- a/tests/learning/test_trial_runner.py
+++ b/tests/learning/test_trial_runner.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from pathlib import Path
 from unittest.mock import MagicMock, patch
 
 from openjarvis.evals.core.types import RunConfig, RunSummary
@@ -137,7 +138,9 @@ class TestBuildRunConfig:
         cfg = runner._build_run_config(trial, recipe)
 
         assert "trial-abc" in cfg.output_path
-        assert cfg.output_path.startswith("out/")
+        # Path(...) / "..." legitimately produces backslashes on Windows,
+        # so compare path components rather than a hard-coded "out/".
+        assert Path(cfg.output_path).parts[0] == "out"
 
     def test_default_model_fallback(self) -> None:
         runner = TrialRunner(benchmark="supergpqa")


MohammedAlkindi:
Ran this on a real Windows 11 machine (Python 3.13.13, Node 24), which is the environment #785 describes — CI's `windows-latest` job aside, that's rarely how this suite gets exercised, so here are the numbers:

**At the PR head (6d147c3):** all three files pass — `pytest tests/evals/core/test_config_split_parsing.py tests/learning/test_trial_runner.py tests/evals/comparison/test_openclaw_runner_contract.py -q` → **27 passed**.

**Fail-before, same machine:** with `origin/main`'s versions of these three test files against the same tree → **7 failed, 20 passed**, and the failure set is exactly the issue's three mechanisms:

- `test_config_split_parsing.py`: 3 failures from the TOML basic-string escape (`C:\Users\...` → invalid `\U` escape in `tomllib`).
- `test_trial_runner.py::test_output_path_contains_trial_id`: the `startswith("out/")` POSIX literal against a backslash path.
- `test_openclaw_runner_contract.py::test_openclaw_runner_parses_real_agent_json_shape`: fails at `assert result.returncode == 0` — on this box Node dies with exit code 134 and empty stderr when handed the two-key environment, so the detail differs from the issue's `ENOENT`-from-`os.tmpdir()` narrative, but the root cause is the same env replacement, and inheriting `os.environ` fixes it here too.

Comment-only review; I can re-run on this machine if the branch changes.

ElliotSlusky (approved):
Maintainer review at the current head: approved. Windows-focused test set: 27 passed. Changes are test-only and remove POSIX/environment assumptions. Repository rules still require a code-owner approval before merge.

PR #815, opened 2026-08-18
fix: use server engine in chat footer

### LINKED ISSUE #791: Chat footer labels every unrecognised model as 'ollama' instead of asking /v1/info
Description: The chat footer derives the engine label from the model id instead of asking the server. In `frontend/src/components/Chat/InputArea.tsx:471-475`:

```ts
const _CLOUD_PREFIXES = ['gpt-', 'o1-', 'o3-', 'o4-', 'claude-', 'gemini-', 'openrouter/', 'MiniMax-', 'chatgpt-'];
const selectedOwner = useAppStore.getState().models.find((m) => m.id === selectedModel)?.owned_by;
const engineLabel = selectedOwner === 'litellm'
  ? 'litellm'
  : _CLOUD_PREFIXES.some(p => selectedModel.startsWith(p)) ? 'cloud' : 'ollama';
```

Any engine whose `/v1/models` ids are short aliases (not `gpt-*`, `claude-*`, `gemini-*`, ...) is therefore labelled `ollama`. The server already knows the answer: `GET /v1/info` returns `{"model": ..., "agent": ..., "engine": ...}`, it is already fetched on mount (`App.tsx` -> `fetchServerInfo`), stored in `serverInfo`, and `ServerInfo.engine` is already typed (`frontend/src/types/index.ts:188-192`).

This is not only a cosmetic label: the same "this is local inference" assumption drives the welcome copy ("Your AI runs locally — private, fast, and always available") and the Local `$0.0000` column of the savings panel, so a user on a non-Ollama engine is shown savings they are not making.

Steps to Reproduce:
1. Configure any engine whose model ids are short aliases and do not match `_CLOUD_PREFIXES` (any engine registering ids like `fast` / `small` / a vendor alias will do).
2. `jarvis serve`, open the web UI, send a message.
3. The footer under the answer reads `ollama - <model> - ...`, while `curl http://127.0.0.1:8000/v1/info` reports the real engine.

Expected Behavior: the footer names the engine that actually served the request, as reported by the server.

Actual Behavior: it reads `ollama` for every engine the prefix list does not recognise.

Environment: OpenJarvis @ 1ca3ec6 (main), Python 3.13.1, Node 22.15, Windows 11. Platform-independent.

Suggested fix: prefer `useAppStore.getState().serverInfo?.engine` when present and keep the prefix heuristic only as a fallback for the case where `/v1/info` has not resolved yet. A local one-hunk patch doing this is available if useful.

igorpimentelgameiro:
## What does this PR do?

Uses the engine reported by `GET /v1/info` through `serverInfo.engine` for the chat footer instead of relying solely on the selected model ID.

The existing model-prefix heuristic remains as a fallback while server information is unavailable. This prevents engines with short or unrecognized model aliases from being incorrectly labeled as `ollama`.

Fixes #791

## How was this tested?

* `cd frontend && npm test`
* `cd frontend && npm run build`

## Checklist

* [x] Frontend tests pass (`npm test`)
* [x] Frontend production build passes (`npm run build`)
* [x] No public API changes
* [x] No new registry component
* [x] No documentation changes required


ElliotSlusky (dismissed):
Maintainer review at the current head: approved. Full TypeScript/Vite production build passed. The footer now uses the server-reported engine with the existing heuristic as fallback. Repository rules still require a code-owner approval before merge.

PR #816, opened 2026-08-18
test(conftest): isolate the whole suite from the real ~/.openjarvis home

### LINKED ISSUE #787: Test suite resolves config paths against the real ~/.openjarvis home
Description: `openjarvis.core.config`'s module-level constants `DEFAULT_CONFIG_DIR`/`DEFAULT_CONFIG_PATH` (src/openjarvis/core/config.py:62-63) are resolved once, at import time, from `paths.get_config_dir()`. Several code paths that use these constants - `TraceStore.db_path`'s dataclass `default_factory`, `AgentConfigEvolver.__init__`'s `mkdir` calls, `LearningOrchestrator.run()`'s `load_config()` call - resolve against the developer's real `~/.openjarvis`, because `tests/conftest.py` only sets `OPENJARVIS_NO_UPDATE_CHECK` inside a fixture (which runs after module import); no `HOME`/`OPENJARVIS_HOME` override exists anywhere before `openjarvis.core.config` is first imported by the test session.

```python
# tests/conftest.py - no env override before this import
from openjarvis.core.config import GpuInfo, HardwareInfo
```
```python
# core/config.py:62-63
DEFAULT_CONFIG_DIR = get_config_dir()
DEFAULT_CONFIG_PATH = get_config_path()
```

Steps to Reproduce:
1. Note the current contents/mtimes of `~/.openjarvis/agent_configs/` (or its absence) before running tests.
2. Run any test that touches `AgentConfigEvolver` or `LearningOrchestrator`, e.g. `pytest tests/learning/ -k evolver`.
3. Inspect `~/.openjarvis/agent_configs/` (and `.history/`) afterward.
4. Observe files stamped with the test run's timestamp now exist under the developer's real home directory, and that `load_config()` calls made during the test picked up the developer's real `[engine] default` setting rather than a sandboxed default.

Expected Behavior: The test suite never reads from or writes to the developer's real `~/.openjarvis`; every test runs against an isolated, temporary config directory.

Actual Behavior: Because `DEFAULT_CONFIG_DIR`/`DEFAULT_CONFIG_PATH` are bound at import time with no prior environment override, tests silently touch the real home directory.

Environment: OpenJarvis @ 1ca3ec6 (main), Python 3.13.1, Windows 11. Platform-independent - reproduces on any machine with a pre-existing `~/.openjarvis`.

Suggested fix: In `tests/conftest.py`, set `OPENJARVIS_HOME` (and clear `XDG_DATA_HOME`/`OPENJARVIS_CONFIG`) to a fresh temporary directory at module import time, before the `from openjarvis.core.config import ...` line - setting the env var only inside a fixture runs too late, since the constants are already bound by the time any fixture executes. An autouse fixture can additionally re-assert the env vars each test and clear `load_config`'s cache before/after each test. A local patch doing this, plus regression tests, is available if useful.

havolline1:
## Summary
Fixes #787: `openjarvis.core.config.DEFAULT_CONFIG_DIR` / `DEFAULT_CONFIG_PATH` are resolved once at import time (the install-script model, where `OPENJARVIS_HOME` is set before the process starts). `tests/conftest.py` imported this module before any environment override was in place, so the very first import of `openjarvis.core.config` anywhere in the test session — transitively triggered by nearly everything — permanently bound those constants, and the ~45 modules that reference them, to the developer's real `~/.openjarvis` for the rest of the run. Tests touching e.g. `AgentConfigEvolver` or `LearningOrchestrator` read/write real files under the developer's actual home directory.

- Set `OPENJARVIS_HOME` to a session-wide isolated temp directory at the very top of `conftest.py`, before the `openjarvis.core.config` import — matching the proposed solution exactly.
- Added an autouse fixture that re-asserts `OPENJARVIS_HOME` (defense against anything changing it without going through `monkeypatch`) and clears `load_config()`'s `functools.lru_cache(maxsize=1)` before and after every test, so a stale config computed under a different test's environment/monkeypatched values can't leak forward.
- Fixed the one real fallout this surfaced: `tests/agents/test_claude_code.py`'s two `_ensure_runner()` tests patched `pathlib.Path.home()` to control the destination path, but that function actually resolves its destination via `get_config_dir()`, which checks `$OPENJARVIS_HOME` first and never reaches the `Path.home()` fallback once it's set globally (as it now always is). Switched those two tests to set `OPENJARVIS_HOME` directly instead, matching what the code under test actually reads — this makes the tests more correct, not just unbroken.

## Test plan
- [x] Added `test_constant_never_resolves_to_the_real_developer_home` (`tests/core/test_config_paths.py`): asserts `DEFAULT_CONFIG_DIR != Path.home() / ".openjarvis"`. Confirmed it fails in a fresh process without the conftest fix (`DEFAULT_CONFIG_DIR` bound to the real `C:/Users/.../.openjarvis`) and passes with it.
- [x] `uv run pytest tests/core/test_config_paths.py` — 16 passed, including the pre-existing `test_constant_matches_resolver_at_import` (proving the "constant == fresh resolver call" invariant still holds under the isolated env).
- [x] **Full-suite diff** (the real verification, given this touches shared infra every test depends on): ran the complete suite (~7800 tests, matching CI's exact `-m "not live and not cloud and not hub"` invocation) twice — once with this change, once on a clean checkout — and diffed the failure sets. The only difference: the two `test_claude_code.py` tests (now fixed, included in this PR) and this PR's own new regression test (fails-then-passes as expected). Zero other regressions anywhere in the suite; all other failures (mostly the pre-existing missing `openjarvis_rust` native extension, Linux-only RAPL sysfs paths, and a couple of unrelated Windows encoding/platform issues) are identical between the two runs.
- [x] `uv run ruff check` on all three changed files — clean


ElliotSlusky:
Pushed 4f83e4d to neutralize ambient OPENJARVIS_CONFIG during tests and remove the session test home at teardown while keeping import-time path constants stable for the session. Validation under an intentionally invalid ambient config: 42 tests passed, temp-home cleanup verified, Ruff passed.

ElliotSlusky (dismissed):

ElliotSlusky:
Pushed `48cd8fc4` to repair both current CI assertions without weakening suite home isolation.

- The metrics test now pins an empty config directory and asserts the route canonical `# no telemetry data` response.
- The named-persona test now derives its expected path from `get_config_dir()`, so it follows the isolated `OPENJARVIS_HOME` instead of the process real home.

Validation:
- Focused failing tests: 2 passed
- Broader server/persona/config-path suite: 39 passed
- `ruff check src/ tests/`: passed
- `ruff format --check src/ tests/`: 1308 files already formatted
- [Post-push CI](https://github.com/open-jarvis/OpenJarvis/actions/runs/32326988728): all five jobs passed

The current fork CI workflow run was approved.

Issue #266, opened 2026-04-20
Frontend not sending Bearer Token


kalenmike:
### Description

The frontend does not appear to be sending an authorization token and so it cannot reach the backend when running 

```bash
uv run jarvis serve
```

I get `401` on all the `/v1/telementry/$1` endpoints and the `/v1/managed-agents`

I can verifiy that the API key is working on the backed using curl, by applying the bearer token.

Searching the frontend it doesn't look like the bearer token is set anywhere

```bash
╰─❯  rg "Authorization"
src-tauri/src/lib.rs
1060:        .header("Authorization", format!("Bearer {}", supabase_key))

src/lib/api.ts
843:          Authorization: `Bearer ${SUPABASE_ANON_KEY}`,
```

I am able to fix the frontend by early exiting the  AuthMiddleware (openjarvis/server/auth_middleware.py)

```python
    async def dispatch(self, request: Request, call_next):  # noqa: ANN001

        return await call_next(request) # temporary fix

        if self._api_key and self._requires_auth(request.url.path):
            auth = request.headers.get("Authorization", "")
            if not auth:
                return JSONResponse(
                    {"detail": "Missing Authorization header"},
                    status_code=401,
                )
            scheme, _, token = auth.partition(" ")
            if scheme.lower() != "bearer" or token != self._api_key:
                return JSONResponse(
                    {"detail": "Invalid API key"},
                    status_code=401,
                )
        return await call_next(request)
```

I did search the docs and have AI search the project. Hopefully I am not missing something silly.

### Steps to Reproduce

1. Run jarvis serve
2. Open web interface
3. No models load, network shows all 401, server shows 401.

### Expected Behavior

200 on network requests. Bearer token sent with API key to the backend.

### Actual Behavior

No bearer token applied to the authenticated requests.

### Operating System

Linux

### Python Version

3.10

### Hardware

NVIDIA GPU

### Engine

None

### Logs / Traceback

```shell
╰─❯  uv run jarvis serve
  Energy: cpu_rapl (rapl)
  Speech: faster-whisper
  Scheduler: active
  Memory:    active
Starting OpenJarvis API server
  Engine: ollama
  Model:  qwen3.5:9b
  Agent:  orchestrator
  URL:    http://0.0.0.0:8001
INFO:     Started server process [2627578]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
[...]

PR #818, opened 2026-08-18
fix(frontend): send Bearer auth header on connectors API calls

havolline1:
## Summary
Found this live while walking through first-time setup with an API key configured (required whenever the server binds to a non-loopback address, which `jarvis start`'s default `0.0.0.0` triggers).

- `frontend/src/lib/connectors-api.ts` called `fetch()` directly instead of the shared `apiFetch()` helper (`lib/api.ts`), so every `/v1/connectors` request silently 401'd against a server started with an API key — while every other API module the UI calls (models, info, savings, managed-agents, approvals, ...) worked fine, since they already go through `apiFetch()`.
- This is the exact bug class **#266** fixed elsewhere (`apiFetch()`'s own comment references it directly: *"the bug in #266 was that direct fetch() calls omitted the header and 401'd"*) — this file was just missed when that fix landed.
- **Confirmed live**, not just by reading the code: with a real `jarvis serve` instance running and an API key configured, `GET /v1/connectors` 401'd from the browser (via the Vite dev proxy) while an identical `curl` request with the same Bearer token succeeded against the backend directly, and every other endpoint the frontend called worked normally — isolating the bug specifically to this file's missing auth header, not a server- or proxy-side issue. The user-facing symptom: the entire **Data Sources** page gets stuck on "Loading sources..." forever whenever an API key is set.
- Switched all six `fetch()` calls (`listConnectors`, `getConnector`, `connectSource`, `disconnectSource`, `getSyncStatus`, `triggerSync`) to `apiFetch()`, which injects the header automatically. The OAuth popup flow (`window.open`) is unaffected — browser navigation can't carry custom headers regardless, and that consent-flow entry point is exempted server-side through a different mechanism.

## Test plan
- [x] Added `connectors-api.auth.test.ts` (mirrors the existing `api.auth.test.ts` pattern for #266): asserts all six functions send `Authorization: Bearer <key>` when a key is configured.
- [x] Confirmed all six fail against the old code (`expected undefined to be 'Bearer sk-local-123'`) and pass after the fix.
- [x] `npx vitest run` (full frontend suite) — 38 passed, no regressions.
- [x] `npx tsc -b --noEmit` — clean.
- [x] Re-verified in the actual running app after the fix: the Data Sources page loads correctly, showing all 24 available connectors plus the one already connected.

### DIFF
--- a/frontend/src/lib/connectors-api.auth.test.ts
+++ b/frontend/src/lib/connectors-api.auth.test.ts
@@ -1,0 +1,111 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+// Regression: every call in connectors-api.ts must go through apiFetch()
+// (not a bare fetch()) so the Bearer auth header is attached when
+// OPENJARVIS_API_KEY is set. Direct fetch() calls silently 401 against an
+// authenticated server -- the same bug class #266 fixed in api.ts, which
+// this file was missed by. Confirmed live: with a real server + API key
+// configured, /v1/connectors 401'd from the browser while every other
+// endpoint (which already went through apiFetch) worked.
+
+const SETTINGS_KEY = 'openjarvis-settings';
+const fetchMock = vi.fn<typeof fetch>();
+
+// Minimal in-memory localStorage stub so the helpers can run under node
+// (no jsdom dependency) -- mirrors api.auth.test.ts.
+class MemoryStorage {
+  private store = new Map<string, string>();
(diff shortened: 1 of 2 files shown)


ElliotSlusky (approved):
Maintainer review at the current head: approved. Connector auth tests: 6 passed; full TypeScript/Vite production build passed. All connector API calls now use the authenticated fetch helper. Repository rules still require a code-owner approval before merge.

PR #819, opened 2026-08-18
fix: reset connector sync checkpoint and purge indexed content on disconnect

havolline1:
## Summary

Disconnecting a connector (`POST /v1/connectors/{id}/disconnect`) only cleared its credentials — it never touched the `SyncEngine` checkpoint or the connector's previously-ingested chunks in `KnowledgeStore`, both keyed only by `connector_id`. Reconnecting the same connector to a *different* underlying data source (e.g. switching which Obsidian vault is connected, or which account is linked) silently misbehaves two ways:

1. **Stale checkpoint**: the next sync resumes from the OLD source's `cursor`/`since` watermark, inflating the reported `items_synced` with stale counts and risking skipped items in the new source whose timestamps predate that watermark.
2. **Orphaned content + doc_id collision**: if the new source produces a `doc_id` that collides with an orphaned chunk left behind from the old source (e.g. two vaults each containing a file at the same relative path, like `Welcome.md`), the ingestion pipeline's duplicate-`doc_id` dedup silently discards the new content in favor of the stale orphan — the user's real data never gets indexed, with no error surfaced anywhere.

Found this while setting up the Obsidian connector locally: pointed it at vault A, then reconnected to vault B, and the UI kept reporting a stale item count from vault A. Digging in showed the deeper issue — vault B's `Welcome.md` was being silently dropped in favor of vault A's orphaned `Welcome.md` chunk.

## Changes

- `SyncEngine.reset_checkpoint(connector_id)` — clears the saved `sync_state` row for a connector.
- `KnowledgeStore.delete_by_source(source)` — deletes all chunks for a given source, returns the count removed.
- `POST /v1/connectors/{id}/disconnect` now calls both (best-effort, logged on failure, doesn't fail the disconnect request).

## Test plan

- [x] `tests/connectors/test_sync_engine.py` — `test_reset_checkpoint_clears_state`, `test_sync_after_reset_does_not_inherit_prior_items_or_since` (verifies a post-reset sync starts with `since=None` and a zeroed item count)
- [x] `tests/connectors/test_store.py` — `test_delete_by_source` (source-scoped deletion, other sources untouched, no-op on unknown source)
- [x] `tests/server/test_connectors_router.py` — `test_disconnect_clears_stale_sync_checkpoint`, `test_disconnect_purges_previously_ingested_content` (end-to-end through the actual endpoint)
- [x] Full suite run clean on top of this branch (only pre-existing `*_live*` tests requiring real network credentials fail, unrelated to this change)
- [x] `ruff check` clean on all changed files


ElliotSlusky:
Pushed c146518 to model connector-to-indexed-source ownership (including Gmail IMAP), preserve shared sources, cooperatively cancel and join active syncs before purge, restore checkpoints if purge fails, and scope all router-created SQLite resources. Added cancellation, shared-ownership, coherent-cleanup, multi-source purge, and close tests. Validation: 46 focused tests passed; Ruff passed.

ElliotSlusky:
Final maintainer review found and closed one additional start/disconnect race in `7a30995`: a sync request that was still reading its baseline could previously publish its worker after disconnect had already finished purging. Per-connector generation tokens now cancel that stale start before a writer is launched; a deterministic concurrent-request regression test covers the exact interleaving.

Updated validation: **47 passed** across connector router, sync engine, and knowledge store; Ruff check/format passed.

ElliotSlusky (dismissed):

PR #820, opened 2026-08-18
fix: generic connector setup UI + non-Google OAuth credential registration

havolline1:
## Summary

Two coupled bugs left 9 of the 25 available connectors unusable through the Data Sources UI: `apple_health`, `apple_music`, `github_notifications`, `google_tasks`, `news_rss`, `oura`, `spotify`, `strava`, and `weather`.

**1. Frontend — empty "+ Add" panel.** `DataSourcesPage.tsx` only rendered connect UI (buttons, inputs, instructions) when a connector had a hardcoded entry with `steps` in the static `SOURCE_CATALOG`. The 9 connectors above have no such entry, so clicking "+ Add" opened a panel with literally nothing in it — no button, no input, no explanation of what to do. Obsidian, Dropbox, Slack, etc. all work because they happen to have catalog entries; anything without one was silently dead.

Adds a `GenericConnectPanel` fallback, used whenever a connector has no catalog `steps`, that branches on `auth_type`:
- `local` → the same "grant access, then connect" button pattern already used for macOS-local sources.
- `token` → a single API-token input.
- `oauth` (default) → fetches the connector's `oauth_setup` info (already returned by the backend, just never consumed by the frontend) to show setup guidance + a link to the provider's developer dashboard, plus Client ID / Client Secret fields.

**2. Backend — non-Google OAuth credential pairs silently rejected.** Even with a working UI, submitting real Spotify/Strava credentials still failed. `_maybe_oauth_client_pair()` only recognized a pasted `client_id:client_secret` string as a credential pair when it matched Google's client ID format (`*.apps.googleusercontent.com`) — every other OAuth provider's pair fell through to the raw-access-token path and failed with `"<Provider> client credentials not configured"`. The generic `save_client_credentials()` / `OAuthProvider` registry this function calls into was already fully provider-agnostic (Spotify and Strava are both registered providers with their own `setup_url`/`setup_hint`) — only this one format-sniffing check was Google-specific.

Generalizes the check: Google keeps its *exact* existing behavior (still requires the domain-suffix match, so nothing changes there), while any other OAuth provider's colon-containing string is now accepted as a credential pair, since a `client_id:client_secret` pair is the only thing this endpoint ever expects for those connectors.

Found this while connecting Spotify locally: the panel was empty, and after fixing the UI, submitting real credentials still failed with the backend error above.

## Changes
- `frontend/src/pages/DataSourcesPage.tsx` — `GenericConnectPanel` fallback + wiring.
- `frontend/src/types/connectors.ts` — `OAuthSetupInfo` type, `oauth_setup` field on `ConnectorInfo`, `"token"` added to the `auth_type` unions.
- `frontend/src/lib/store.ts` — `auth_type` added to `CachedConnector`.
- `src/openjarvis/server/connectors_router.py` — generalized `_maybe_oauth_client_pair()`.

## Test plan
- [x] `tests/connectors/test_connectors_router_oauth.py` — new parametrized tests for Spotify/Strava credential-pair registration + malformed-pair 400, plus a guard test confirming Google's exact existing behavior is unchanged.
- [x] `frontend`: `tsc -b` clean, `vitest run` clean (32 passed).
- [x] Manually verified in-browser against a running instance: Apple Health (local), Weather (token), and Spotify (oauth, showing the real setup hint + redirect URI from the backend) all now render a working connect panel where they previously rendered nothing. Submitted a real Spotify `client_id:client_secret` pair end-to-end and confirmed `oauth_required` is returned (was a 400 before this fix).
- [x] Full suite run clean on top of this branch; only pre-existing `*_live*` tests requiring real network credentials fail (unrelated), and one pre-existing test-fixture-ordering flake unrelated to this change (`test_connect_client_pair_returns_oauth_required_no_browser[gdrive]`, reproduces identically on `main` without this branch's changes).
- [x] `ruff check` clean on all changed backend files.


ElliotSlusky:
Repaired against current `main` and updated the fork head without force-pushing.

Validated: `14 passed` for router endpoint coverage; `18 passed` connector coverage; `npm run build`. The generic panel now persists GitHub/Oura tokens, requires and stores Weather location, and requires validated News/RSS feed URLs.

ElliotSlusky (dismissed):
Reviewed repaired connector configuration flow and targeted evidence.

ElliotSlusky:
Follow-up 5cb89cc0d4a755afc8c8e2078c8314c69aed2186 fixes the full-CI lint failure: all six E501 violations and the unused import are resolved, including the previously unreported formatting drift in the OAuth connector test. Full ruff check src/ tests/ and ruff format --check src/ tests/ now pass.

ElliotSlusky (dismissed):
Full-repository Ruff follow-up validated after the functional repair.

PR #821, opened 2026-08-18
fix(cli): write init config and memory files as UTF-8

### LINKED ISSUE #769: jarvis init --digest crashes with UnicodeEncodeError on Windows
Description: `jarvis init`'s config/memory file reads and writes omit `encoding=`, so Python falls back to the locale codec - `cp1252` on a default Windows install - instead of UTF-8. The `--digest` path additionally embeds a literal Unicode box-drawing character (U+2500) in the digest section text, so `target.write_text(existing + digest_section)` raises `UnicodeEncodeError` before the command finishes, leaving `config.toml` without the digest section and no success message printed.

```python
digest_section = """
# --- Morning Digest ---  (contains a literal U+2500 box-drawing rule in the actual source)
...
"""
...
existing = target.read_text()
target.write_text(existing + digest_section)
```

Steps to Reproduce:
1. On a Windows machine with the default `cp1252` console/locale codec (i.e. `PYTHONUTF8` and `PYTHONIOENCODING` unset, and no UTF-8 code page active).
2. Run `jarvis init --digest`.
3. Observe an unhandled `UnicodeEncodeError: 'charmap' codec can't encode character '\u2500'` and `config.toml` left without the digest section.

Expected Behavior: `jarvis init --digest` completes successfully regardless of the OS locale codec.

Actual Behavior: The command crashes on Windows with the default (non-UTF-8) locale codec.

Environment: OpenJarvis @ 1ca3ec6 (main), Python 3.13.1, Windows 11. Windows-only - reproduces specifically when the process's locale codec is not UTF-8, which is the Windows default outside of an explicit UTF-8 code page.

Suggested fix: Add `encoding="utf-8"` to every `read_text()`/`write_text()` call touched by `init()` (the `--config` read/write, the main config write, the digest section read/append/read-back, and the SOUL.md/MEMORY.md/USER.md scaffold writes), and replace the box-drawing character with a plain-ASCII divider so the digest text itself doesn't depend on the write path passing an explicit encoding. A local patch doing this, plus a regression test, is available if useful.

MohammedAlkindi:
Fixes #769

**Summary**

`jarvis init`'s config and memory-file reads/writes omitted `encoding=`, so Python fell back to the locale codec — cp1252 on a default Windows install. The digest section text embeds a literal U+2500 box-drawing rule, so `jarvis init --digest` crashed with `UnicodeEncodeError: 'charmap' codec can't encode character` before printing the success message, leaving `config.toml` without the digest section.

**Changes**

- `src/openjarvis/cli/init_cmd.py`: added `encoding="utf-8"` to all nine bare `read_text()`/`write_text()` calls in the init path (config round-trip, digest append, SOUL.md / MEMORY.md / USER.md creation). No behavior change on platforms whose locale is already UTF-8.
- `tests/cli/test_init_digest.py`: regression test running `init --engine ollama --digest --no-download` through `CliRunner` (same harness as `test_init_host.py`), asserting exit code 0 and that the U+2500 rule survives an explicit-UTF-8 read.

**Testing**

On Windows 11 / Python 3.13 (locale cp1252 — the failing environment):

- Fail-before: the new test against `main`'s `init_cmd.py` fails with exactly the reported error — `UnicodeEncodeError('charmap', '# OpenJarvis configuration…', 467, 470, 'character maps to <undefined>')` surfacing as a non-zero exit code.
- Pass-after: `pytest tests/cli/test_init_digest.py tests/cli/test_init_host.py tests/cli/test_init_guidance.py tests/cli/test_quickstart.py` → **32 passed**.
- `ruff check` clean on both changed files.

Note: on a UTF-8-locale system the old code happens to work, so the regression test only has teeth on the `windows-latest` CI job (or any non-UTF-8 locale) — stating that so the coverage claim is honest.

Scope note: #769 only names the `init` path and that is all this touches; other modules with bare `read_text/write_text` are deliberately left alone.


ElliotSlusky (dismissed):
Maintainer review at the current head: approved. Focused CLI suite: 32 passed; Ruff passed. All relevant init reads and writes now use explicit UTF-8. Repository rules still require a code-owner approval before merge.

MohammedAlkindi:
2f45f09d auto-dismissed the approval, and it does need a real re-read rather than a rubber stamp. Upstream changed `--config` to select an existing file to install rather than an alternate output path, so both write sites now target `DEFAULT_CONFIG_PATH`. Upstream had already added `encoding="utf-8"` to the config write, so that hunk is upstream verbatim. The digest read-modify-write below it kept the locale default, which is the crash this branch fixes, so explicit encoding stays on both the reads and the write.

`.github/CODEOWNERS` is `* @jonsaadfalcon @ANarayan @robbym-dev`, so a code-owner approval is what unblocks merge, which matches the note in the dismissed review. @jonsaadfalcon could you take it?

Issue #426, opened 2026-05-28
Unsupported value: 'temperature' does not support 0.7 with this model


dv336699:
### Description

Brand new install, never installed jarvis before and getting an error.

### Steps to Reproduce

curl -fsSL https://open-jarvis.github.io/OpenJarvis/install.sh | bash
jarvis

### Expected Behavior

To work

### Actual Behavior

Error: Error code: 400 - {'error': {'message': "Unsupported value: 'temperature' does not support 0.7 with this model. Only the default (1) value is supported.", 'type':
'invalid_request_error', 'param': 'temperature', 'code': 'unsupported_value'}}

### Operating System

macOS

### Python Version

3.11

### Hardware

Apple Silicon

### Engine

None

### Logs / Traceback

```shell
curl -fsSL https://open-jarvis.github.io/OpenJarvis/install.sh | bash
jarvis
OpenJarvis installer
  install dir: /Users/dv336699/.openjarvis
  WSL2:        0

[..] Install uv
    uv already installed
[ok] Install uv
[..] Clone OpenJarvis repo
Cloning into '/Users/dv336699/.openjarvis/src'...
remote: Enumerating objects: 2164, done.
remote: Counting objects: 100% (2164/2164), done.
remote: Compressing objects: 100% (1917/1917), done.
remote: Total 2164 (delta 272), reused 1323 (delta 205), pack-reused 0 (from 0)
Receiving objects: 100% (2164/2164), 36.41 MiB | 13.39 MiB/s, done.
Resolving deltas: 100% (272/272), done.
[ok] Clone OpenJarvis repo
[..] Copy install scripts
[ok] Copy install scripts
[..] Create venv
Using CPython 3.11.15
Creating virtual environment at: .openjarvis/.venv
Activate with: source .openjarvis/.venv/bin/activate
[ok] Create venv
[..] Install OpenJarvis
Using Python 3.11.15 environment at: /Users/dv336699/.openjarvis/.venv
Resolved 65 packages in 785ms
      Built openjarvis @ file:///Users/dv336699/.openjarvis/src
Prepared 27 packages in 3.39s
Installed 65 packages in 540ms
 + aiohappyeyeballs==2.6.2
 + aiohttp==3.13.5
 + aiosignal==1.4.0
 + annotated-doc==0.0.4
 + annotated-types==0.7.0
 + anyio==4.13.0
 + attrs==26.1.0
 + backoff==2.2.1
 + brotli==1.2.0
[...]

PR #822, opened 2026-08-18
fix: complete OAuth connect flow for Spotify/Strava + register Claude 5 model family

havolline1:
## Summary

**Note:** this branch is stacked on #820 (not yet merged) — the `GenericConnectPanel` frontend fallback from that PR is a direct prerequisite for actually reaching Spotify's connect UI at all, so this diff includes #820's changes too until it merges. Happy to rebase once #820 lands if that's preferred.

Two independent sets of fixes, both found and verified while getting a real Spotify connection and a real Claude API connection working end-to-end on a local install:

### 1. OAuth connect flow for non-Google providers (Spotify, Strava)

Connecting Spotify surfaced a chain of four coupled bugs that made the OAuth-based connect flow for any non-Google provider completely unreachable, even with correct credentials:

- `SpotifyConnector.is_connected()` / `StravaConnector.is_connected()` only checked that the credentials file existed, not that it contained a real access token. The file is written as soon as a Client ID/Secret pair is registered — before the user completes OAuth consent — so the UI's popup-polling flow saw "connected" immediately and never opened the consent screen. The connector then reported connected with no access token, and the first sync crashed with `KeyError: 'access_token'`.
- `AuthMiddleware` gated every `/v1/*` path uniformly, including `oauth/start` and `oauth/callback` — both exist specifically to be hit by a plain top-level browser navigation (a popup `window.open` target, and the provider's own redirect back), which can never carry a custom Authorization header. On any server with an API key configured (required for non-loopback binds), the whole flow 401'd on "Missing Authorization header" before it could start.
- The redirect_uri shown to users (via the static, provider-specific `setup_hint`) didn't match what `/oauth/start` actually sent (built dynamically from the request's own `base_url`). Registering the documented URI with the provider still failed with "redirect_uri: Not matching configuration".
- The dynamically-built redirect_uri used whatever host the browser reached the API through ("localhost"), which Spotify's policy (enforced for all apps created after April 2025) rejects outright — it requires the explicit loopback IP `127.0.0.1`.

### 2. Claude 5 model family not recognized + temperature rejected

- `CloudEngine`'s static `_ANTHROPIC_MODELS` list had never been updated past the 4.x generation, so configuring `default_model = "claude-sonnet-5"` with a valid `ANTHROPIC_API_KEY` silently fell back to a local Ollama model — indistinguishable from an actual reachability problem, since the fallback is silent by design.
- Once routed to the real API, the Claude 5 family also rejects a non-default `temperature` outright (`"`temperature` is deprecated for this model"`), unlike every prior generation. Generalizes the existing OpenAI-specific unsupported-temperature retry (#426) to also match Anthropic's error message, applied at all three Anthropic call sites (generate, stream, stream_full).

## Test plan

- [x] `tests/connectors/test_spotify.py`, `tests/connectors/test_strava.py` — `is_connected()` requires a real access token, not just file existence.
- [x] `tests/server/test_auth_middleware.py` — `oauth/start`/`oauth/callback` exempt from auth; every other `/v1/connectors/*` route still requires it.
- [x] `tests/server/test_connectors_router.py` — `oauth_setup.redirect_uri` matches exactly what `/oauth/start` sends, including `localhost` → `127.0.0.1` normalization.
- [x] `tests/engine/test_cloud_extended.py`, `tests/engine/test_cloud_stream_full.py` — Claude 5 family registered; temperature-rejection retry covers both the non-streaming and streaming (manual `__enter__`/`__exit__`) paths, confirmed to genuinely fail without the fix via a temporary revert.
- [x] Frontend: `tsc -b` clean, `vitest run` clean (32 passed).
- [x] Manually verified end-to-end against real services: created a real Spotify Developer app, connected it through the UI, confirmed a successful sync with no errors; sent a real chat completion to `claude-sonnet-5` through a running server and got back real model output.
- [x] Full suite clean on top of this branch; only pre-existing failures unrelated to this change (`*_live*` tests needing real network credentials, `test_gemma_cpp.py` needing local model weights, and one pre-existing connector-registry test-ordering flake reproducible identically on unmodified `main`).
- [x] `ruff check` clean on all changed backend files.


ElliotSlusky:
Reconciled as a non-force merge on top of the repaired #820 line and current `main`.

Security/behavior validation: `88 passed` across OAuth router, auth middleware, Spotify/Strava, and CloudEngine suites; `npm run build`. Public callbacks now use one-time cryptographic OAuth state, trusted callback-base configuration, and escaped error HTML. Verified current official Anthropic source: only `claude-sonnet-5` is registered, at its active introductory $2/$10 per MTok rate through 2026-08-31.

ElliotSlusky (dismissed):
Reviewed stacked OAuth security repair and targeted evidence.

ElliotSlusky:
Pricing verification source: [Anthropic’s Sonnet 5 announcement](https://www.anthropic.com/research/claude-sonnet-5) states the active introductory API price is $2 input / $10 output per MTok through 2026-08-31, then $3/$15. The repair therefore retains only the documented `claude-sonnet-5` ID and the currently active rate.

ElliotSlusky:
Follow-up c3ca8cd73d4de8df9b18bcb75073a99e3c2c7d69 merges #820 lint repair 5cb89cc and fixes the additional OAuth-specific Ruff violations. Full ruff check src/ tests/ and ruff format --check src/ tests/ now pass on the stacked branch; the prior 88 OAuth/auth/cloud tests and frontend build remain behaviorally unchanged.

ElliotSlusky (dismissed):
Full-repository Ruff follow-up validated after the functional repair.

ElliotSlusky:
Follow-up 043ec8f0b2713abbcbe30c6036c6bf7bf3719b5e addresses the Linux full-suite failure. The legacy redirect-URI integration test now supplies the trusted callback-base configuration required by the OAuth hardening instead of relying on TestClient synthetic host testserver, which is intentionally rejected. Validation: 20 router/OAuth tests passed; full Ruff check/format passed.

ElliotSlusky (approved):
Trusted callback-base integration follow-up verified.

ElliotSlusky:
Closing as stale/bundled after current-main reconciliation. The generic connector setup portion was already repaired and merged through #820. The remaining branch combines OAuth security changes with unrelated Claude model/pricing changes and still requires a fresh, authenticated OAuth-initiation design. Please resubmit the OAuth flow and Claude support as separate focused PRs based on current main. The source branch has not been deleted.

PR #824, opened 2026-08-19
Integrate Mixedbread retrieval for deep research capabilities

krypticmouse:
## What does this PR do?

Integrates [Mixedbread's toast-1 search agent](https://www.mixedbread.com/blog/toast-1) as an **opt-in** retrieval option, at two layers:

**1. `mixedbread` memory backend** (`tools/storage/mixedbread_backend.py`)
A `MemoryBackend` registered in `MemoryRegistry`: `store()` uploads to a Mixedbread store, `retrieve()` runs agentic (toast-1) search. Selected via `[tools.storage] default_backend = "mixedbread"`; guarded import + `memory-mixedbread` extra, so the SDK stays optional. Thread-safe store resolution (lock + 409-conflict retry keyed for cross-process races).

**2. Deep Research retrieval** (`connectors/mixedbread_search.py`)
`MixedbreadSearch` is duck-type compatible with `HybridSearch` for the research loop: toast-1 does query decomposition/ranking server-side, while the **local SQLite knowledge base stays canonical** — remote hits map back to local chunks via `external_id`, so citations, deep links, participants, thread context, and person/time/source filters behave exactly as with hybrid search. `MixedbreadKnowledgeSync` mirrors `knowledge_chunks` idempotently (`jarvis mixedbread-sync`, with cloud-upload confirmation). Any API failure falls back to local hybrid search.

**Opt-in by construction**: the default path never imports mixedbread code — `build_research_search()` lives in `hybrid_search.py` and only imports the cloud module inside the `[deep_research] retrieval = "mixedbread"` branch (enforced by a test that blocks the module). Default config = `"hybrid"`, unchanged behavior.

Also adds `scripts/bench_retrieval_ab.py` + `scripts/prep_bcp_subset.py`, a reproducible retrieval A/B harness over a BrowseComp-Plus subset.

## How was this tested?

- **35 new offline unit tests** (fake-client, no network): backend contract, hydration/filter semantics, sync idempotency, fallback behavior, concurrency races (regression tests for the store-creation race), and default-path isolation from the SDK.
- **Offline stress harness**: 32-way concurrent first-use (1 store created, was 32 before the lock), 16×100 mixed ops, 5MB/unicode/NUL payloads, fault injection, clear/store interleaving — all pass.
- **Live end-to-end** against the real API: memory backend (8/8 checks: upload latency p50 ~500ms, searchable in ~3s, delete/clear semantics) and Deep Research path (8/8: multi-hop query joining two docs in 3.2s, thread context, source/person filters, CLI, store cleanup).
- **Retrieval A/B on BrowseComp-Plus** (20 queries, 424 docs = labeled evidence + hard negatives, identical chunks both sides, Ollama embeddings at 99% coverage):

| arm | evidence recall@10 | MRR@10 | gold-doc@10 | p50 latency |
|---|---|---|---|---|
| hybrid (local BM25+vector) | 0.368 | 0.440 | 0.65 | 0.38s |
| mxbai (managed, non-agentic) | 0.559 | 0.746 | 0.80 | 0.44s |
| **toast-1 (agentic)** | **0.926** | **1.000** | **1.00** | 8.50s |

toast-1 ranked a labeled evidence doc #1 on all 20 queries; the trade is ~8.5s/query vs ~0.4s local — hence opt-in for research-grade questions, local hybrid remains the default.

## Checklist

- [x] Tests pass (`uv run pytest tests/ -v`) — pre-existing exceptions: live-connector tests need Oura/Strava/Spotify/Google credentials; `test_rust_bridge` needs the native extension built
- [x] Follows registry pattern (`MemoryRegistry.register("mixedbread")`)
- [x] Documentation updated (`docs/user-guide/memory.md`, `docs/user-guide/deep-research.md`, config template)


ElliotSlusky:
Repair pushed in `17758812aaa4683f2e9ca29d3799622217cb7b48` after an append-only merge of current `main` (`759bfd0b98e9f0d391aa8482ca7b3c7dbd5bceea`).

What changed:
- sync now tags its own remote files, walks every Mixedbread file-list page, and deletes only stale mirror-owned files; unrelated direct uploads are preserved
- local soft-deleted chunks are excluded from hydration and removed remotely on the next explicit sync
- Mixedbread adapters now close SDK clients they own, and the CLI/server research paths release search/store resources
- cloud sync/search use public KnowledgeStore and HybridSearch APIs instead of reaching into private SQLite/filter internals
- the lockfile was regenerated from current main with `mixedbread==0.54.0`

Verification:
- `58 passed` across hybrid search, KnowledgeStore, Mixedbread search/sync, and Mixedbread memory tests
- Ruff check and format check pass on all touched Python files
- modified modules compile with `py_compile`

Focused coverage includes paginated stale deletion, preservation of non-mirror files, deleted-row filtering, person/time/source filtering, and owned-versus-injected client lifecycle. No merge was performed.

PR #826, opened 2026-08-19
Optimize DenseMemory delete path

rpreslar4765:
## Summary
- replace the O(n) matrix reallocation on delete with a soft-delete tombstone + compaction threshold
- keep retrieval and count logic focused on active documents only
- add a regression test covering delete semantics

## Why
Each `delete()` call previously rebuilt the NumPy matrix by deleting a row, which turns repeated deletions into quadratic work as the index grows. This is especially visible in long-lived dense memory stores that churn entries.

## Validation
- `PYTHONPATH=src python -m pytest tests/tools/storage/test_dense.py -q`


ElliotSlusky:
Repair pushed in `1b1d45f15fc864691d9883f93544ddec6b744de8` after an append-only merge of current `main` (`759bfd0b98e9f0d391aa8482ca7b3c7dbd5bceea`).

The outstanding review concerns are addressed:
- retrieval computes scores against the bounded physical matrix and masks tombstoned scores in the new 1-D score vector, avoiding an O(n × dimension) advanced-index matrix copy on every query
- the 25% compaction trigger is now a named threshold
- regression coverage now exercises threshold compaction, the all-deleted reset path, tombstone filtering without row indexing, idempotent deletion, and concurrent delete/retrieve consistency

Verification:
- full DenseMemory test file: `22 passed, 9 skipped` (the skips require a live Ollama embedding model)
- Ruff check and format check pass
- modified modules compile with `py_compile`

No merge was performed.

PR #831, opened 2026-08-19
fix: KokoroTTSBackend device selection is dead code

kire610:
## Summary

`KokoroTTSBackend.__init__` accepts a `device` parameter (default `"auto"`) and stores it, but `_ensure_pipeline()` never actually forwards it to `KPipeline(...)`:

```python
self._pipeline = KPipeline(lang_code="a")
```

`KPipeline`'s own `device` parameter defaults to `None`, which means "auto-detect CUDA, else CPU" (`kokoro/pipeline.py`). Since `device` was never passed at all, this class's setting had no effect either way — on a CUDA-capable machine it happens to end up on the GPU regardless of what's configured, purely because auto-detect picks CUDA when available.

The real-world impact shows up in the opposite case: explicitly requesting `device="cpu"` — for example to leave VRAM headroom for an LLM sharing the same GPU, a real constraint on 8GB-class cards — is silently ignored; it lands on CUDA anyway.

Separately, even if the stored value had been passed through as-is, the literal string `"auto"` isn't something `KPipeline` understands (it checks `device == 'cuda'` or `device is None`, not `"auto"`), so passing it verbatim would have hit `.to("auto")` and raised.

## Fix

Translate this class's `"auto"`/`""` sentinel to `KPipeline`'s own `None` sentinel, and forward the result:

```python
device = None if self._device in ("", "auto") else self._device
self._pipeline = KPipeline(lang_code="a", device=device)
```

## Test plan

Verified on an RTX 5070 (CUDA 13, torch 2.13.0+cu130):

- [x] `pipeline.model.device` reports `cuda:0` with the default `device="auto"`
- [x] A real `synthesize("Hello from Jarvis.")` call produced valid 24kHz audio (90KB for the test sentence)
- [ ] Not tested: explicit `device="cpu"` forcing CPU-only inference on a CUDA-capable machine (the actual bug this fixes) — I don't have an easy way to verify device placement without a GPU present to compare against; would appreciate a maintainer double-checking that path if there's an existing test fixture for it.

Small, self-contained fix — happy to add a unit test if there's a preferred pattern for mocking `kokoro.KPipeline` in this repo's test suite.

### DIFF
--- a/src/openjarvis/speech/kokoro_tts.py
+++ b/src/openjarvis/speech/kokoro_tts.py
@@ -30,7 +30,12 @@ class KokoroTTSBackend(TTSBackend):
         try:
             from kokoro import KPipeline
 
-            self._pipeline = KPipeline(lang_code="a")
+            # KPipeline treats device=None as "auto-detect CUDA, else CPU"
+            # (kokoro/pipeline.py) -- it does NOT understand the literal
+            # string "auto" this class defaults to, so that has to be
+            # translated here rather than passed through directly.
+            device = None if self._device in ("", "auto") else self._device
+            self._pipeline = KPipeline(lang_code="a", device=device)
         except ImportError:
             raise RuntimeError(
                 "kokoro package not installed. Install with: pip install kokoro"


ElliotSlusky:
Closing as superseded. Current main already contains the repaired Kokoro implementation: shared KModel ownership, configured model_path/device handling, multilingual pipelines, and removal of the unsupported Korean language code. Applying this stale one-line branch would regress that newer implementation. The source branch has not been deleted.

PR #832, opened 2026-08-19
Add App Launcher tool — open/close installed desktop apps by name

ninjabbk1:
## What does this PR do?

Adds a new `app_launcher` tool (`src/openjarvis/tools/app_launcher.py`,
`@ToolRegistry.register("app_launcher")`) that lets an agent open or close
already-installed desktop applications by free-text name — e.g. "abra o
Chrome", "abra a calculadora", "feche o Spotify" — reusing the existing
tool-routing system (it's just a tool call, not a new module/agent) and
mirroring `shell_exec.py`'s `code:execute` capability pattern.

**Discovery** (Windows): reads the `App Paths` registry key
(`HKLM`/`HKCU`) and resolves Start Menu `.lnk` shortcuts via the Windows
Shell COM API (`pywin32`, new optional `app-launcher` extra, Windows-only
marker) — never a hardcoded install path. A small nickname table
normalizes generic terms ("navegador" → default browser, "calculadora" →
"calc") before falling back to fuzzy matching (`difflib`) for typos.

**Matching behavior** (see `find_app`/`_match_precise`/`_match_fuzzy`):
raw query term tried first (exact → substring), then alias-expansion
terms only if that finds nothing, then fuzzy matching only against the
user's own raw term (never against aliases — an alias is already a
precise, curated name, and fuzzy-matching it produced false positives
like "calculadora" ~ "CapCut" during testing). Candidates sharing the
same display name but different paths (e.g. two "Discord" shortcuts)
collapse to one match; candidates with genuinely different names (e.g.
"Google Chrome" vs "Google Chrome Canary") correctly trigger
disambiguation, asking which one the user meant.

**Safety** (explicit requirements from the brief): never installs new
software (no such action exists), never runs a caller-supplied path (the
tool's only input is a free-text `app_name` — every launchable path comes
from OS-level discovery of already-installed programs), never requests
admin elevation (plain `os.startfile`/unelevated `Popen`), and never
silently claims success — a not-found app or a failed launch always
returns `success=False` with a clear message.

**Closing** apps uses `tasklist`/`taskkill` (graceful terminate, escalating
to a forced kill only if that doesn't work) — kept dependency-free
(no `psutil`) since Windows ships these tools already.

## How was this tested?

- 22 new unit tests (`tests/tools/test_app_launcher.py`) covering: opening
  a found app, the not-found warning, disambiguation with multiple close
  matches, alias resolution, missing/unknown-action error handling,
  closing a running process, and confirming the tool's parameters never
  expose a raw path (only `action`/`app_name`).
- Full `tests/tools/` regression suite: 687 passed, 0 failures.
- **Manually verified end-to-end against real installed software** on a
  Windows machine: discovered 125 real installed apps via the registry +
  Start Menu, correctly resolved "bloco de notas"/"paint"/"discord"/
  "opera"/"winrar" to their real executables, correctly reported
  "chrome"/"spotify" as not installed (honest, not fabricated) on that
  machine, and actually opened + closed Notepad through the tool
  (confirmed via `tasklist` before/after).
- `ruff check` / `ruff format --check` pass on all changed files.

## Checklist

- [x] Tests pass (`uv run pytest tests/tools/ -v`)
- [x] Follows registry pattern (`@ToolRegistry.register`)
- [ ] Documentation updated (not yet — flagging for a follow-up if merged)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,6 +93,7 @@ desktop = [
     "faster-whisper>=1.0",
 ]
 openhands = ["openhands-sdk>=1.0; python_version >= '3.12'"]
+app-launcher = ["pywin32>=306; sys_platform == 'win32'"]
 gpu-metrics = ["pynvml>=12.0"]
 energy-amd = ["amdsmi>=6.1"]
 energy-apple = ["zeus-ml[apple]"]
--- a/src/openjarvis/tools/__init__.py
+++ b/src/openjarvis/tools/__init__.py
@@ -152,4 +152,9 @@ try:
 except ImportError:
     pass
 
+try:
+    import openjarvis.tools.app_launcher  # noqa: F401
+except ImportError:
+    pass
+
 __all__ = ["BaseTool", "ToolExecutor", "ToolSpec"]
--- a/src/openjarvis/tools/app_launcher.py
+++ b/src/openjarvis/tools/app_launcher.py
@@ -1,0 +1,642 @@
+"""App Launcher tool — open/close already-installed desktop applications by name.
+
+Discovers installed programs from the OS's own registries of installed
+software (Windows: the ``App Paths`` registry key and Start Menu shortcuts)
+rather than depending on a hardcoded path. Deliberately narrow in scope:
+
+- Never installs new software (no "install" action exists at all).
+- Never runs a file the caller supplies directly — the only input is a free-
+  text ``app_name``; every launchable path comes from OS-level discovery of
+  *already installed* programs, never from an arbitrary caller-provided path.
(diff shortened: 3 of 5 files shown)


ElliotSlusky:
Closing for redesign rather than merging this stale security-sensitive implementation. Opening applications and terminating process trees must require explicit confirmation and a narrowly defined allowlist; the current tool can kill applications without confirmation, exposes executable paths, and is not covered by the repository data-boundary classifications. Please resubmit a focused version with confirmation, provenance/allowlist policy, audit events, and Windows CI tests. The source branch has not been deleted.

PR #833, opened 2026-08-19
Add jarvis desktop — native app window instead of a browser tab

ninjabbk1:
## What does this PR do?

Adds `jarvis desktop`, a new CLI command that opens the existing web UI in
a native OS application window — its own title bar and icon, no address
bar, no tabs, no browser chrome — instead of the user having to open a
browser tab at `localhost:PORT`. The frontend (already served statically
by `jarvis serve`, see `server/app.py`'s static-file mount) and the
backend are both completely untouched; only how the UI is *displayed*
changes.

**Why `pywebview` instead of finishing the existing Tauri app** in
`frontend/src-tauri/` (which is real, but partial/never compiled in this
environment): building it would require installing the full Rust + MSVC
build toolchain — a heavy, slow, one-time setup with real risk of
Windows-specific build issues. `pywebview` needs a single lightweight
Python package (using the OS's already-installed WebView2 runtime on
Windows) and fully satisfies every stated requirement — a native window
with zero browser controls around the same running web UI — without
compiling anything. If/when the project wants the full native Tauri
build, this doesn't preclude that; it's a pragmatic, low-risk way to get
the actual user-visible outcome today.

**How it works** (`src/openjarvis/cli/desktop_cmd.py`):
1. Spawns `jarvis serve --host 127.0.0.1` as a plain (non-detached) child
   process — deliberately not detached like `jarvis start`'s daemon spawn
   in `cli/daemon_cmd.py`, since this process's lifecycle must stay tied
   to the window so it can be terminated cleanly.
2. Polls `GET /health` until the backend is ready (60s timeout).
3. Opens a `pywebview` window titled "OpenJarvis (J.A.R.V.I.S.)", using
   the Tauri app's already-committed icon
   (`frontend/src-tauri/icons/icon.ico` — reused rather than duplicated;
   falls back to no icon if that path isn't present, e.g. a built wheel).
4. On window close (or any exception, or interpreter exit via `atexit`),
   always terminates the backend child process — escalating from
   `terminate()` to `kill()` if it doesn't exit within 5s — so nothing is
   left running in the background.

Also updates the desktop shortcut's launcher script to call
`jarvis desktop` instead of the old "open a browser tab" flow (that
script isn't part of this repo — it's a local file on the machine this
was built on).

## How was this tested?

- 9 new unit tests (`tests/cli/test_desktop_cmd.py`) covering icon
  resolution, health-polling success/timeout/non-200 paths, and
  subprocess cleanup (`_stop_server` against both a live and an
  already-exited process) — the GUI-opening part itself isn't
  unit-testable without a display, so:
- **Manually verified live, end-to-end, with a real window**: wrote a
  driver script that starts the real backend, opens the real window
  pointed at it, waits 5s (confirming the UI actually renders), then
  programmatically closes the window (`webview.windows[0].destroy()`,
  simulating a user clicking the close button) and asserts the backend
  subprocess is confirmed dead afterward via `proc.poll()`. Output:
  `server pid alive AFTER cleanup: False`.
- Also tested the worst case — force-killing the top-level wrapper
  process (not a graceful window close) — and confirmed via `tasklist`
  that zero `python.exe` processes and no listening backend were left
  behind either way.
- `tests/cli/` full regression: 484 passed, 8 pre-existing failures
  (verified identical on a clean `main` with none of this PR's changes —
  memory-backend/scan tests unrelated to this change, needing the Rust
  extension / a Linux-only assumption).
- `ruff check` / `ruff format --check` pass on all changed files.

## Checklist

- [x] Tests pass (`uv run pytest tests/cli/test_desktop_cmd.py -v`)
- [x] Follows registry pattern (Click command registered in `cli/__init__.py`, same as every other subcommand)
- [ ] Documentation updated (not yet — flagging for a follow-up if merged)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,6 +93,7 @@ desktop = [
     "faster-whisper>=1.0",
 ]
 openhands = ["openhands-sdk>=1.0; python_version >= '3.12'"]
+desktop-window = ["pywebview>=5.0"]
 gpu-metrics = ["pynvml>=12.0"]
 energy-amd = ["amdsmi>=6.1"]
 energy-apple = ["zeus-ml[apple]"]
--- a/src/openjarvis/cli/__init__.py
+++ b/src/openjarvis/cli/__init__.py
@@ -98,6 +98,7 @@ if not _DATA_BOUNDARY_BOOTSTRAP:
     from openjarvis.cli.config_cmd import config
     from openjarvis.cli.connect_cmd import connect
     from openjarvis.cli.daemon_cmd import restart, start, status, stop
+    from openjarvis.cli.desktop_cmd import desktop
     from openjarvis.cli.digest_cmd import digest
     from openjarvis.cli.doctor_cmd import doctor
     from openjarvis.cli.eval_cmd import eval_group
@@ -126,6 +127,7 @@ if not _DATA_BOUNDARY_BOOTSTRAP:
     cli.add_command(ask, "ask")
     cli.add_command(chat, "chat")
     cli.add_command(serve, "serve")
+    cli.add_command(desktop, "desktop")
     cli.add_command(model, "model")
     cli.add_command(memory, "memory")
     cli.add_command(mine, "mine")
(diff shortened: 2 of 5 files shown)


ElliotSlusky:
Closing because this introduces a parallel pywebview desktop shell alongside the established Tauri application, duplicating backend-child lifecycle and expanding the dependency/attack surface without GUI CI. The current implementation also accepts arbitrary hosts and can attach to an unrelated service that merely answers the health endpoint. Please improve the existing Tauri desktop path or open a design discussion before proposing a second shell. The source branch has not been deleted.

PR #883, opened 2026-08-21
fix: make managed schedules deterministic

### LINKED ISSUE #775: Managed agent ticks always resolve to hardcoded gemma4:31b model
Description: The `or` chain resolving the tick model in `AgentExecutor` puts the hardcoded `_AGENT_TICK_DEFAULT_MODEL = "gemma4:31b"` constant before `self._system.model` (src/openjarvis/agents/executor.py:291-297). Since the constant is unconditionally truthy, any managed agent tick without an explicit per-agent `config["model"]` always requests `gemma4:31b`, regardless of the system's actually configured default model - making `self._system.model` and the `if not model: raise FatalError` guard both dead code. If `gemma4:31b` isn't pulled locally, the engine's "model not found" error is classified fatal and the tick fails with no retry.

```python
model = (
    config.get("model")
    or _AGENT_TICK_DEFAULT_MODEL
    or (self._system.model if self._system else "")
)
if not model:
    raise FatalError("No model configured for agent")
```

Steps to Reproduce:
1. Create a managed agent with no `config["model"]` set.
2. Set `self._system.model` to some other model (e.g. `"qwen3:8b"`).
3. Invoke a tick and inspect the model actually resolved/requested.
4. Observe it is always `"gemma4:31b"`, never the system's configured model.

Expected Behavior: The resolution order is per-agent config, then the system's configured default, then the hardcoded constant as a last resort.

Actual Behavior: The hardcoded constant always wins whenever no per-agent `model` is set, regardless of the system default.

Suggested fix: Reorder to `config.get("model") or (self._system.model if self._system else "") or _AGENT_TICK_DEFAULT_MODEL`, making the hardcoded constant a true last-resort fallback. A local patch doing this, plus regression tests, is available if useful.

### LINKED ISSUE #776: Cron-scheduled tasks and managed agents ignore the configured timezone
Description: Two independent schedulers evaluate cron expressions without honoring a configured timezone. `TaskScheduler._compute_next_run`/`_compute_next_cron` (src/openjarvis/scheduler/scheduler.py:283-303) always evaluates against `datetime.now(timezone.utc)`, so a task's `metadata["timezone"]` (set by `proactive_agent.py`/`morning_digest.py`) is stored but never consulted - a `'0 5 * * *'` schedule always fires at 05:00 UTC instead of 5am in the intended zone. `_next_cron_fire` in `src/openjarvis/agents/scheduler.py` (lines 19-36), the separate scheduler used for managed agents, has the same gap but is worse: it uses naive local time (`datetime.datetime.fromtimestamp(base)`) with no timezone parameter at all.

```python
# scheduler/scheduler.py
def _compute_next_run(self, task: ScheduledTask) -> Optional[str]:
    now = datetime.now(timezone.utc)
    ...
    if task.schedule_type == "cron":
        return self._compute_next_cron(task.schedule_value, now)
```
```python
# agents/scheduler.py
def _next_cron_fire(cron_expr: str, now: float | None = None) -> float:
    ...
    base = now or time.time()
    dt = datetime.datetime.fromtimestamp(base)
    cron = croniter(cron_expr, dt)
    return cron.get_next(datetime.datetime).timestamp()
```

Steps to Reproduce:
1. Create a scheduled task with `schedule_type="cron"`, `schedule_value="0 5 * * *"`, and `metadata={"timezone": "America/Los_Angeles"}`.
2. Call `TaskScheduler._compute_next_run(task)` (or run the scheduler with the system clock set to a known UTC time) and check the resulting `next_run`.
3. Observe the computed time is 05:00 UTC, not 05:00 Los Angeles time - the configured timezone metadata is never read.

Expected Behavior: A cron schedule with a configured timezone fires at the intended local time in that zone.

Actual Behavior: The timezone is stored but has no effect; schedules always evaluate in UTC (`scheduler/scheduler.py`) or naive local time (`agents/scheduler.py`).

Suggested fix: Have both cron computation functions accept an optional IANA timezone name, convert `now` into that zone via `zoneinfo.ZoneInfo` before handing it to `croniter`, and (for `TaskScheduler`, which stores/compares UTC ISO strings) convert the result back to UTC before returning it. `TaskScheduler` can read `task.metadata.get("timezone")` automatically; `AgentScheduler` needs a new `config["timezone"]` key threaded through `register_agent`. A local patch doing this, plus regression tests, is available if useful.

### LINKED ISSUE #777: croniter is optional; missing it silently degrades cron to hourly
Description: `croniter` is declared only under `[project.optional-dependencies] scheduler` in `pyproject.toml` (line 128), not in the base `dependencies` list, so a default `uv sync`/`pip install openjarvis` leaves it absent. `_next_cron_fire` in `src/openjarvis/agents/scheduler.py` (lines 19-30) catches the resulting `ImportError` and silently degrades any cron schedule to a flat 3600-second interval, logging only a single `logger.warning` - so, for example, a "weekday mornings at 8am" schedule fires hourly around the clock forever, with no loud, actionable signal that anything is wrong.

```python
def _next_cron_fire(cron_expr: str, now: float | None = None) -> float:
    try:
        from croniter import croniter
    except ImportError:
        logger.warning("croniter not installed, treating cron as 3600s interval")
        return (now or time.time()) + 3600
    ...
```

Steps to Reproduce:
1. `pip install openjarvis` (base package, no `scheduler` extra) into a clean environment.
2. `python -c "import croniter"` - observe `ModuleNotFoundError`.
3. Register a managed agent with a cron schedule like `"0 8 * * 1-5"` and run the scheduler.
4. Observe the agent fires roughly every hour, 24/7, instead of only weekday mornings at 8am, with the mis-scheduling visible only as a single `logger.warning` line.

Expected Behavior: Cron parsing is always available (croniter is a hard dependency), or a missing croniter fails loudly at schedule-registration time rather than silently degrading the schedule's actual behavior.

Actual Behavior: A missing optional dependency silently and permanently changes a cron schedule's semantics with only a warning-level log line.

Suggested fix: Move `croniter` into the base `[project] dependencies` so cron parsing is always available, and have `_next_cron_fire` raise loudly (e.g. `RuntimeError`) if it is somehow still missing at runtime, rather than falling back to an hourly approximation. A local patch doing both, plus regression tests, is available if useful.

### LINKED ISSUE #944: AgentExecutor builds RouterPolicy with available_models=[model], so routing is always a no-op
### Description

`AgentExecutor` is the only runtime caller of a `RouterPolicy`, but it constructs the policy with a **single-element** `available_models` list containing the model that has already been selected. A policy can therefore only ever return that same model, making the router a no-op wherever it is enabled.

### Where

`agents/executor.py`:

```python
model = (
    config.get("model")
    or _AGENT_TICK_DEFAULT_MODEL
    or (self._system.model if self._system else "")
)
...
router_policy_key = config.get("router_policy")
if router_policy_key and self._system:
    try:
        from openjarvis.core.registry import RouterPolicyRegistry
        from openjarvis.learning.routing.router import build_routing_context

        policy = RouterPolicyRegistry.create(
            router_policy_key,
            available_models=[model],      # <-- only the already-chosen model
        )
        instruction = config.get("instruction", "")
        ctx = build_routing_context(instruction)
        selected = policy.select_model(ctx)
        if selected:
            model = selected
    except Exception:
        pass  # Fall back to configured model
```

Every branch of `HeuristicRouter.select_model` resolves through `_smallest_model`, `_largest_model`, `_find_model_by_tag` or a default -- all of which return the sole entry when the list has length 1. `selected` is therefore always equal to `model`, and the assignment is a no-op.

### Expected Behavior

The policy should receive the set of models the system could actually serve (from the discovered engines / `ModelRegistry`), so it can choose between them.

### Actual Behavior

Setting `router_policy` on an agent has no observable effect.

### Secondary concern

The bare `except Exception: pass` means a genuine failure inside routing -- an unregistered policy key, an import error, a malformed context -- is indistinguishable from routing having run and chosen the same model. Debug-level logging of the exception would make misconfiguration visible instead of silent.

### Related

- #942 -- `_largest_model` ranks by `parameter_count_b`, which is `0.0` for every cloud model, so cloud can never be selected even when the candidate list is correct.
- The `ask.py` issue filed alongside this one -- the primary entry point performs no routing at all.
- #775 touches the same block (`_AGENT_TICK_DEFAULT_MODEL` resolution).

Together these mean the documented routing feature is non-functional through every path: the main CLI never calls it, the one caller that does can't choose, and the ranking function can't rank the intended targets.

### Version

1.0.4.dev210+gc1af3c94 -- Windows 11, Python 3.12.10

---

*Found while setting up local-first routing with cloud escalation; investigated with Claude Code.*

ElliotSlusky:
## Summary
- resolve managed-agent models from agent config, then the live system default, with the legacy model only as a last resort
- evaluate task and managed-agent cron expressions in their configured IANA timezone and persist task times in UTC
- install croniter by default and fail loudly if cron support is unavailable instead of silently converting cron jobs into hourly jobs

## Validation
- `pytest tests/agents/test_executor.py tests/agents/test_scheduler.py tests/scheduler/test_scheduler.py -q` — 50 passed
- `pytest tests/agents tests/scheduler -q -m "not live and not cloud and not hub"` — 629 passed, 1 skipped, 1 unrelated native loop-guard failure that reproduces unchanged on current main with the same installed extension
- `uv lock --check` passed with uv 0.12.5

Fixes #775
Fixes #776
Fixes #777

The current implementation also supplies distinct available engine models to routing policies instead of a singleton candidate list. Fixes #944.


ElliotSlusky:
Repair appended at 8abbc022.

- Managed-agent routing now gives the selected policy every distinct model reported by the active engine. The resolved default stays first when it is actually available; discovery failure safely falls back to the resolved model.
- Morning-digest scheduling now threads the configured IANA timezone through both CLI and server paths and persists it in task metadata, so cron evaluation occurs at the requested local wall time.
- Adversarial coverage verifies multi-model code routing, excludes an unavailable legacy fallback when real models exist, preserves remote fallback behavior, and confirms a persisted 06:00 Los Angeles task converts back to 06:00 local time.

- 63 focused executor, managed-scheduler, task-scheduler, digest-schedule, and digest CLI tests passed.
- Ruff check and format verification passed on all affected files.

PR #885, opened 2026-08-21
fix(memory): preserve short paragraphs during chunking

### LINKED ISSUE #754: chunk_text() silently drops short paragraphs below min_chunk_size
Description: In `chunk_text()`, the mid-document flush (src/openjarvis/tools/storage/chunking.py:78-92) only appends the accumulated buffer to `chunks` when it reaches `min_chunk_size` (default 50 tokens), but always resets `current_tokens` afterward regardless of whether it was appended. When the buffer is below `min_chunk_size` and also below `chunk_overlap` (so the overlap-carry branch doesn't fire either), that text is discarded with no trace. The identical pattern exists in the oversized-paragraph pre-flush a few lines below.

```python
if current_tokens and len(current_tokens) + len(para_tokens) > cfg.chunk_size:
    chunk_content = " ".join(current_tokens)
    if _count_tokens(chunk_content) >= cfg.min_chunk_size:
        chunks.append(Chunk(content=chunk_content, source=source,
                             offset=chunk_start_offset, index=len(chunks)))
    if cfg.chunk_overlap > 0 and len(current_tokens) > cfg.chunk_overlap:
        current_tokens = list(current_tokens[-cfg.chunk_overlap:])
    else:
        current_tokens = []
```

Steps to Reproduce:
1. `from openjarvis.tools.storage.chunking import chunk_text`
2. `text = "IMPORTANT my API key rotation policy is documented here" + "\n\n" + " ".join(f"word{i}" for i in range(600))`
3. `chunks = chunk_text(text)` and check whether any chunk contains `"IMPORTANT"`.
4. Observe 2 chunks are produced and neither contains the leading short paragraph.

Expected Behavior: Every non-empty paragraph appears in at least one chunk, even a short one that falls below `min_chunk_size` before the document has produced its first chunk.

Actual Behavior: A short paragraph that occurs before a later flush point is silently dropped from the output; `jarvis memory index` reports success ("Indexed 2 chunks") with the lead-in content missing from the index.

Suggested fix: When the mid-document (or oversized-paragraph) flush skips emitting a sub-floor buffer, carry it forward into the next chunk's accumulation instead of discarding it, adjusting the tracked offset accordingly. A local patch doing this, plus regression tests, is available if useful.

ElliotSlusky:
## Summary
- retain below-floor paragraphs until they can be emitted with following content
- carry short lead-ins into oversized paragraph windows without losing tokens
- add exact oversized and normal-paragraph regressions

## Tests
- 14 tests in tests/memory/test_chunking.py
- Ruff check and format verification on changed files

Fixes #754

The final chunk is now retained even below the preferred minimum size, so short final paragraphs and oversized-paragraph tails remain searchable. Added offset and complete-coverage regressions across paragraph boundaries and overlaps. Validation on current main: 86 chunking, storage-tool, and memory-CLI tests passed, plus Ruff and formatting.


ElliotSlusky:
Appended the chunk-size blocker fix in `7c5842f`.

Reproduction on `1b4854f`: with `chunk_size=10`, `min_chunk_size=5`, and no overlap, a 3-token lead-in followed by a 10-token paragraph produced one 13-token chunk. The chunker now streams paragraphs into bounded windows, preserves paragraph boundaries when possible, carries sub-floor lead-ins without dropping them, rebalances short tails when feasible, and bounds retained overlap so it always makes forward progress.

Validation:
- 19 focused chunking tests passed, including adversarial lead-in/overlap/minimum combinations
- an additional deterministic sweep of 10,000 configurations/text layouts verified every emitted chunk is non-empty, at most `chunk_size`, and sequentially indexed
- Ruff check/format clean
- broader memory/storage run reached 196 passed / 39 skipped; its 45 failures all require the unavailable `openjarvis_rust` extension (local native build is blocked by the host's unaccepted Xcode license), so hosted CI is authoritative for those backends

ElliotSlusky:
Repair appended at 330740ca.

- ChunkConfig now rejects non-integer/bool/zero/negative chunk sizes and negative/non-integer overlaps, and chunk_text revalidates mutable configs before entering its loop.
- CLI ranges reject chunk_size below 1 and chunk_overlap below 0 before indexing.
- The memory-index tool advertises JSON Schema minimums and fails invalid direct calls without storing anything.
- A deterministic 88-configuration boundary sweep verifies forward progress, non-empty hard-bounded chunks, complete token coverage, sequential indexes, and offsets that map back to the original flattened document.

- 53 focused chunking and storage-tool tests passed.
- The focused CLI rejection regression passed.
- Direct adversarial probes for chunk_size 0/-1 and overlap -1 now fail immediately with ValueError.
- Ruff check and format verification passed.

PR #889, opened 2026-08-21
fix: validate proactive SMS drafts

### LINKED ISSUE #835: sms_draft_reply silently drops the SMS body and reports success
### Description

The proactive-agent prompt requires all SMS action payloads to use contact and body.

Location: src/openjarvis/agents/proactive_agent.py:88

```
- For sms actions: contact (phone/email), body (the message text)
```

However, the sms_draft_reply executor reads draft instead of body.

Location: src/openjarvis/tools/proactive_tools.py:417-419

```python
if atype == "sms_draft_reply":
    # Draft only — surface in next digest, don't send
    return True, f"Draft saved: {payload.get('draft', '')[:80]}"
```

A valid model-generated payload such as:

```
{
  "contact": "+15551234567",
  "body": "Sounds good, see you tomorrow."
}
```

therefore produces:

```
Draft saved:
```

The action still reports success and is marked as executed, although the body is ignored and no draft is actually saved or surfaced in a later digest.

### Steps to Reproduce

This issue was identified by an automated testing tool for academic research and manually verified. If you have any concerns about this type of reporting, please let me know, and I will adjust my workflow accordingly.

### Expected Behavior

The executor should use the documented body field.

### Actual Behavior

The action still reports success and is marked as executed, although the body is ignored and no draft is actually saved or surfaced in a later digest.

### Operating System

Linux

### Python Version

3.12

### Hardware

NVIDIA GPU

### Engine

None

### Logs / Traceback

```shell

```

ElliotSlusky:
## Summary
- read SMS draft text from the documented `body` field and require a contact plus nonblank body
- surface the validated draft without sending it
- mark only successful pending actions executed and report attempted/executed/failed totals accurately

## Validation
- focused proactive tests — 8 passed
- broader tools/proactive sweep — 683 passed, 15 optional-dependency skips

Fixes #835


ElliotSlusky:
Repair appended at 4283bad0.

- SMS draft contact/body are accepted only when each value is an actual nonblank string; nulls, numbers, collections, and whitespace are rejected instead of being stringified.
- Invalid drafts are rejected before entering the approval queue, leaving no pending or approved row.
- Defense-in-depth execution validation also prevents legacy malformed approved rows from being reported or persisted as executed.

- 12 focused proactive-action tests passed, including queue and execution probes for null, non-string, and whitespace payloads.
- Ruff check and format verification passed.

PR #892, opened 2026-08-21
fix: require consent before uninstalling data

### LINKED ISSUE #772: jarvis-uninstall.sh deletes all user data with no prompt or backup
Description: `scripts/install/jarvis-uninstall.sh` runs `rm -rf "$OPENJARVIS_HOME"` unconditionally, with no confirmation prompt, no `--dry-run`, and no backup (lines 27-29). Because the install root and the data root are the same directory (`~/.openjarvis`), this instantly and irrecoverably destroys `config.toml`, `SOUL.md`/`MEMORY.md`/`USER.md`, `skills/`, `scheduler.db`, `telemetry.db`, and `connectors/*.json` OAuth refresh tokens. The script's header comment only lists directories being removed and gives no hint that personal data is included.

```bash
if [[ -d "$OPENJARVIS_HOME" ]]; then
    rm -rf "$OPENJARVIS_HOME"
    echo "Removed $OPENJARVIS_HOME"
fi
```

Steps to Reproduce:
1. Have a populated `~/.openjarvis` (config, stored memory facts, connected OAuth tokens).
2. Run `scripts/install/jarvis-uninstall.sh` (or `jarvis-uninstall` if installed as a shim).
3. Observe the entire directory is deleted immediately with no prompt of any kind and no way to recover it.

Expected Behavior: The uninstaller warns what will be deleted (config, memory, skills, databases, OAuth tokens) and requires explicit confirmation before deleting, with a non-interactive opt-out flag for scripted use.

Actual Behavior: It deletes everything unconditionally the moment it runs.

Environment: OpenJarvis @ 1ca3ec6 (main), bash, Windows 11 (WSL/Git Bash) or any POSIX shell - platform-independent, the script is POSIX shell and applies to any OS it runs on.

Suggested fix: Add a confirmation prompt that lists exactly what will be deleted before the `rm -rf`, gated so it only fires when `$OPENJARVIS_HOME` exists, plus a `-y`/`--yes` flag to skip the prompt for scripted/non-interactive use; declining aborts with no deletion. A local patch doing this is available if useful.

ElliotSlusky:
## Summary
- warn exactly which config, memory, skill, database, and connector credential data will be deleted
- require the user to type `yes`, with `-y`/`--yes` as the explicit automation opt-out
- abort without deleting data or shims when confirmation is declined
- refuse install roots resolving to `/` or the user home

## Validation
- full Python installer suite — 60 passed
- shell syntax check passed

Fixes #772


ElliotSlusky:
Appended the destructive-root blocker fix in `224e1d5`.

Reproduction on `779e00d` used a real temporary sandbox with `HOME=/tmp/.../users/alice` and `OPENJARVIS_HOME=/tmp/.../users`: `--yes` exited 0 and recursively deleted the entire `users` directory, including a sibling `bob/sibling-data` sentinel.

The uninstaller now rejects the user's home and every ancestor of it, rejects known broad system/temp roots, and requires a real non-symlinked installer state file containing a known completed OpenJarvis install stage. Recursive deletion uses the exact physical path that passed those checks. `--yes` remains supported for both the default and a valid custom owned root.

Validation:
- 100 installer/deployment tests passed
- adversarial subprocess tests exercise an ancestor target with a sibling sentinel, unowned custom data, and a symlinked fake ownership marker
- `bash -n` clean
- Ruff check/format clean for the Python tests

PR #896, opened 2026-08-21
fix: require desktop inference consent

### LINKED ISSUE #274: Allow inference source during setup of Desktop App
### Problem Statement

I have no intention of allowing the desktop app to automatically install any software, especially not Ollama. 

Image

### Proposed Solution

Allow us to configure the backend before startup instead of automatically trying to install Ollama.

### Primitive Area

Frontend

### Alternatives Considered

_No response_

ElliotSlusky:
## Summary

- keep fresh desktop installs inert until the user explicitly chooses an inference source
- require a separate, clearly disclosed confirmation before Ollama can start or download a model
- let users configure an OpenAI-compatible endpoint before backend startup, with safe Back/Cancel paths
- persist the selected source before starting, while preserving automatic startup for returning users with an explicit saved config
- reject missing, malformed, and implicit `{}` configs on both the app-launch and `start_backend` paths

Fixes #274

## Validation

- `npm test` — 59 passed
- `npx tsc --noEmit`
- `npm run build`
- `cargo test` — 35 passed (using the CommandLineTools compiler/SDK directly because this host has the Xcode GUI license prompt pending)
- `git diff --check`


ElliotSlusky:
Appended the setup-recovery blocker fix in `9d2fbd5`.

The prior flow persisted a first-run source before asynchronous boot, treated `start_backend` task creation as success, and exposed no way out of the progress/error screen. A bad custom URL or failed Ollama/model startup therefore survived relaunch and replayed the same failure.

The recovery contract is now transactional:
- first-run choices are staged as unconfirmed and become confirmed only after the complete backend reaches `ready`
- every failed/aborted pending boot rolls its config back; an interrupted pending choice is also discarded on relaunch, while legacy and previously confirmed returning-user configs remain trusted
- boot tasks are tracked and abortable, so “Change inference source” cancels endpoint/model work before child processes are stopped
- the setup screen exposes that action throughout startup and on errors; reset stops children, clears the choice, resets status, and returns to the inert chooser
- an immediate `start_backend` rejection also rolls back the staged choice

Validation:
- frontend: 61 tests passed
- TypeScript + Tauri production build passed
- Rust: 37 tests passed, including pending-config rollback and tracked-task cancellation
- `cargo fmt --check` and `git diff --check` clean
- Rust tests used the standalone CommandLineTools compiler/SDK because the GUI Xcode license remains pending on this host

ElliotSlusky:
Repair appended at `00eae1d833f6038015efdd2260086b95c4e00ab9`.

Security/lifecycle repair:
- first-run custom API keys are staged in a dedicated OS secure-storage slot and injected only into the managed child process;
- the durable `<ENGINE>_API_KEY` is promoted only after all readiness gates pass;
- cancellation, reset, startup failure, and config-write failure discard the staged key while preserving any pre-existing provider key;
- cloud-key hot reload now requires both `server_ready` and a live `jarvis` child owned by this desktop process, so a responsive port-8000 squatter never receives raw credentials;
- first-run setup with a staged key refuses to attach to an unowned healthy loopback server.

Verification:
- `cargo test` — 47 passed (includes stage/promote, failed-write restoration, cancel, reset, child-abort, and port-squatter regressions)
- `cargo check --tests` — passed
- `cargo fmt --check` — passed
- `npm test` — 61 passed
- `npm run build` — passed
- `npm run build:tauri` — passed
- `git diff --check` — passed

PR #827, opened 2026-08-19
fix: wire capability/rate-limit/audit enforcement into server + managed-agent paths

kire610:
The configured capability policy now reaches server, managed-agent, CLI, SDK, MCP, scheduler and learning execution paths with a stable caller identity. Shared/server profiles enable default-deny enforcement, and privileged tools and direct execution paths are checked before I/O. Rate-limit and denial events are included in the serialized audit chain.

This closes the remaining failure modes from #836: explicitly enabled capability enforcement cannot silently disappear when initialization fails; missing or invalid policy files stop startup; supplying only a policy path preserves the profile's enforcement defaults; canonical tool capabilities cannot be weakened by a ToolSpec; anonymous callers and explicitly empty per-agent policies cannot inherit default grants. Explicit user overrides remain honored. Personal profile capability enforcement remains opt-in. The exact baseline grants and caller identities are documented in the security guide.

The branch preserves the original PR history and incorporates current main plus #932's audit serialization, retaining the scheduling and stream lifecycle fixes already merged.

Validation:

- 1,889 passed and 7 optional cryptography skips across security, all server/agent/SDK/MCP tests and adjacent CLI/tool/learning paths.
- 58 final policy/config/factory tests passed, including the final explicitly empty agent policy regression.
- Ruff lint and format checks pass for the full source and test trees (1,370 files); git diff --check passes.
- Tests include denied/allowed real tool execution, managed and HTTP server paths, policy startup errors, direct execution/search gates, stable identities, rate-limit attribution and concurrent audit-chain verification. No live provider calls were required.

Fixes #836

PR #897, opened 2026-08-21
feat: add native dynamic weather tool

### LINKED ISSUE #392: native weather tool with OpenWeatherMap integration
### Problem Statement

OpenJarvis currently has no native weather tool. Users are forced to use shell_exec workarounds which are not dynamic, require manual API key injection, and have hardcoded city values.

A native weather tool should:
- Read API key from config.toml or environment variable
- Accept dynamic city/location from the user's query
- Return structured weather data (temperature, description, humidity, wind)
- Support metric and imperial units
- Support multiple languages
- Work seamlessly with all agents (simple, native_react, orchestrator)

Proposed config:
[tools.weather]
provider = "openweathermap"
api_key = "${OPENWEATHERMAP_API_KEY}"
default_location = "auto"
units = "metric"
lang = "de"

Example queries that should work:
- "What is the weather in Vienna?"
- "Aktuelles Wetter in Sankt Johann in Tirol"
- "Will it rain tomorrow in Innsbruck?"

### Proposed Solution

Add a native weather tool (src/openjarvis/tools/weather.py) that:
- Reads OPENWEATHERMAP_API_KEY from environment or config.toml
- Extracts city/location dynamically from the user query
- Returns structured weather data in the user's language
- Registers as a standard tool like web_search or shell_exec
- Works with all agents without shell_exec workarounds

### Primitive Area

Tools

### Alternatives Considered

shell_exec workaround with hardcoded city and manual API key injection —
not dynamic, not production-ready, hallucination risk when key missing.

ElliotSlusky:
OpenJarvis agents can now call `get_weather` for a dynamic city or city/country location, metric or imperial units, localized current conditions, and up to 120 hours of three-hour forecasts. The tool works through the shared tool resolver used by simple, native_react, and orchestrator agents. Non-secret defaults live in `[tools.weather]`; credentials come from the tool credentials store, `OPENWEATHERMAP_API_KEY`, or an existing Weather connector.

Weather requests use fixed OpenWeatherMap endpoints, validate arguments before network access, and declare their external data-boundary and `network:fetch` capability. Provider errors and HTTPX request logs redact API keys, including when INFO logging is enabled. Forecast responses are capped even if the provider ignores the requested count, and malformed connector and tool credential entries are ignored with safe connector fallback. The tool picker reports credential availability without exposing the key.

Reconciled the branch with current main, preserving the original PR history and both web-search credential providers. Weather registration now uses the literal registry key so the capability inventory audit recognizes it.

Validation:
- 287 focused weather, connector, configuration, credential, agent resolver, server, tool-registration, and data-boundary tests passed.
- Fourteen real TOML credential regressions cover numeric, list, boolean, and table values plus malformed sections, both without credentials and with a valid connector fallback; they preserve the file and prevent invalid network calls.
- Real HTTPX MockTransport checks verify credential-safe logging on both success and HTTP errors, unchanged outbound authentication, fixed endpoint selection, query encoding, and unaffected unrelated request logging.
- Ruff, full formatting check (1,365 files), and `git diff --check` passed.
- Documentation build passed.
- Full default suite before the final two-line malformed-tool-credential guard, in a freshly synced locked Python 3.12 environment with the native extension: 8226 passed, 62 skipped, 272 warnings in 159.63s (0:02:39). Live/cloud/hub tests were excluded.

Live OpenWeatherMap API access was not tested because no API credential was provided.

Fixes #392


ElliotSlusky:
Repair appended at `eaa17915c20dc4a646d5d41f5f64efc8c72b8d85`.

Reload-stability repair:
- `WeatherTool` now resolves the connector class, fetch function, and `WeatherAPIError` through the live connector module instead of stale imported identities;
- weather tests use the same module-qualified access;
- added a deterministic regression that clears/repopulates `ConnectorRegistry`, then verifies the actionable sanitized provider error is still caught and the API key is not exposed.

Verification:
- focused weather/connector/router/config/CLI/agent/server suite — 88 passed
- deterministic serial `tests/server/test_connectors_router.py tests/tools/test_weather.py` — 48 passed on #897
- synthetic #898 then #897 stack, same deterministic serial suite — 52 passed (the prior 3 failures are gone)
- Ruff check + format check — passed
- `git diff --check` — passed

Tanergurbuz369:
While testing a local build that merges this PR alongside #827 (capability gate fix), I found an integration hazard between the two that's worth flagging before either merges:

`WeatherTool` is registered with `@ToolRegistry.register(_TOOL_NAME)` (a module-level variable), here:

```python
_TOOL_NAME = "get_weather"
...
@ToolRegistry.register(_TOOL_NAME)
class WeatherTool(BaseTool):
```

#827 introduces `tests/security/test_capabilities.py::TestCapabilityPolicy::test_every_registered_builtin_has_an_explicit_inventory_entry`, which discovers registered tools by parsing source with `ast` and only recognizes `ToolRegistry.register("literal_string")` calls — a variable argument like `_TOOL_NAME` isn't picked up by that scan. Every other tool in the codebase registers with a literal string, so this one is the outlier.

Net effect: `get_weather` silently isn't counted as "registered" by that test's AST scan, so `DEFAULT_TOOL_CAPABILITIES` (in #827) has no matching entry for it and the new inventory test fails once both PRs land — even though the tool itself works fine at runtime.

Suggested fix, whichever PR merges second:

```python
@ToolRegistry.register("get_weather")
class WeatherTool(BaseTool):
```

(`_TOOL_NAME` can stay for the other internal uses — `tool_id`, `spec.name`, credential lookup — just the decorator argument needs to be the literal.)

Verified this both reproduces (test fails with both PRs merged) and this one-line change fixes it, in a disposable local branch — not proposing a PR against either of your branches, just flagging the interaction since neither PR alone would surface it in CI.

PR #935, opened 2026-09-02
Add voice interruption support to Jarvis GUI

MikeyWorldwide:
## Summary
- Ctrl+Space or the mic button now interrupts Jarvis mid-speech: kills TTS playback immediately, transitions to listening, and begins accepting the new command (natural conversational barge-in) instead of being ignored while speaking.
- `VoiceCore` now tracks a speak token so a stale utterance can never start/resume playback after an interruption, even if the interrupt lands while pyttsx3 is still synthesizing rather than already playing.

## Test plan
- [x] Real-audio unit test of `VoiceCore.speak()`/`stop_speaking()`: interrupting mid-playback stops audio within ~10ms and `speak()` returns immediately; interrupting mid-synthesis (before any audio exists) is also caught and never plays; a subsequent `speak()` call after interruption plays normally.
- [x] Full-app test driving the real `JarvisGUI` (real `mainloop()`, real pyttsx3/sounddevice output): while speaking the mic button reads "Interrupt"; simulated Ctrl+Space mid-sentence flips state to "listening" immediately, stops playback, and the next spoken command is actually submitted; state returns to "idle" afterward.
- [x] Confirmed `<Control-space>` is bound via `bind_all`.

https://claude.ai/code/session_01QLqyypKq3ftvP98PxL6g9q

### DIFF
--- a/jarvis_gui.py
+++ b/jarvis_gui.py
@@ -1,0 +1,409 @@
+"""JARVIS COMMAND CENTER - desktop GUI.
+
+Maps onto the command-center architecture:
+  GUI          -> this file (customtkinter)
+  JARVIS CORE  -> jarvis_core.JarvisCore (HTTP client for the OpenJarvis
+                  backend's router/providers/memory/telemetry on :8000)
+  VOICE CORE   -> jarvis_voice.VoiceCore (local STT via faster-whisper,
+                  local TTS via pyttsx3+sounddevice, push-to-talk - no
+                  wake-word listening yet)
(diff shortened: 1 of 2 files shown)


MikeyWorldwide:
@jonsaadfalcon @ANarayan @robbym-dev friendly ping for review when you get a chance — this adds voice interruption support (Ctrl+Space / mic button barge-in) to the Jarvis GUI. Let me know if anything needs adjusting.

PR #975, opened 2026-09-09
fix(security): hard-gate cloud engine behind intelligence.allow_cloud

cynarAI:
## Summary
- `intelligence.provider = "local"` in `config.toml` was purely decorative — no code path ever checked it. Four independent places could reach a real cloud API (OpenAI/Anthropic/Google/OpenRouter) just from an env var being present or from the request body, regardless of the configured provider.
- Adds `IntelligenceConfig.allow_cloud: bool = False` and gates all four paths on it:
  - `cli/serve.py` — no longer builds a `CloudEngine` at startup unless `allow_cloud` is true, even if cloud API key env vars are set.
  - `engine/_discovery.py` — `get_engine()`/`discover_engines()` never select/construct a cloud-classed engine (`is_cloud=True`) as a fallback or as an explicit request when `allow_cloud` is false.
  - `server/routes.py` `POST /v1/cloud/reload` — now returns `403` outright when `allow_cloud` is false, instead of writing submitted keys to `os.environ` and hot-wiring a `CloudEngine` into the live `MultiEngine`.
  - `server/routes.py` `_uses_direct_cloud_router()` — no longer routes a chat request to the real cloud API based on the model name looking cloud-shaped (`gpt-*`, `claude-*`, ...) when `allow_cloud` is false.

## Test plan
- [x] `uv run --extra dev pytest tests/engine/test_discovery.py tests/engine/test_cloud.py tests/engine/test_cloud_extended.py tests/cli/test_serve_single_build.py -q` → 82/82 passed
- [x] Live-verified on the running `jarvis-serve.service`: local chat completions unaffected (`kat-coder-v2.5` responds normally), `/v1/cloud/reload` with a test key → `403`, `get_engine()` never returns the cloud engine even with a fake `OPENAI_API_KEY` in the environment (explicit request and fallback discovery both refuse), flipping `allow_cloud=true` restores the prior behavior with no regression.

### DIFF
--- a/configs/openjarvis/prompts/jarvis-constitution.md
+++ b/configs/openjarvis/prompts/jarvis-constitution.md
@@ -1,0 +1,37 @@
+# Jarvis-Konstitution
+
+Einzige Quelle für Ton, Verhalten und Autonomie-Grenzen über alle Kanäle (Telegram, apex, openjarvis-Desktop). Wird nicht automatisch überschrieben — Änderungen laufen über denselben Dev-Workflow wie jede andere Code-Änderung (PR, Review durch den Self-Audit, Merge nach "go" vom User).
+
+## Ton
+
+- Direkt, keine Beruhigungswörter, kein Smalltalk.
+- Ergebnisse und Entscheidungen zuerst nennen, nicht drumherum reden.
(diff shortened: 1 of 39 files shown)


cynarAI:
Opened by mistake against the wrong repo (should have targeted my own fork) — closing, no action needed here.

PR #1007, opened 2026-09-18
fix(cli): report Ollama pull streaming errors

johnnycwatt:
## What does this PR do?

Fixes `jarvis model pull --engine ollama` reporting success when Ollama sends an error during a streamed download.

Previously, `ollama_pull()` checked the HTTP status but ignored JSON frames containing an `error` field. An HTTP 200 response containing progress followed by `{"error":"disk full"}` therefore printed `Successfully pulled ...` and caused the command to exit with status 0.

The helper now detects an error frame, displays the diagnostic as literal text, and returns `False`. The existing command-level handling then exits with status 1. Literal rendering preserves bracket-containing diagnostics without interpreting them as Rich markup. The response remains managed by its existing context manager.

[Ollama's error documentation](https://docs.ollama.com/api/errors#errors-that-occur-while-streaming) explains that errors occurring after streaming begins are sent as newline-delimited JSON objects containing an `error` property. The HTTP status cannot change once the response has started, so `raise_for_status()` alone cannot detect these failures. See also the [model pull endpoint documentation](https://docs.ollama.com/api/pull).

The adjacent OpenJarvis server implementation requests `stream=False`. Ollama can wait for that download's outcome before choosing the HTTP status; its non-streaming `waitForStream()` path converts errors into HTTP error responses. This differs from the CLI's streaming path, which must inspect error frames in the response body. The distinction is visible in Ollama's [`PullHandler`, `waitForStream`, and `streamResponse` implementations](https://github.com/ollama/ollama/blob/main/server/routes.go).

This change is limited to explicit error frames. It preserves successful downloads, the helper's Boolean interface, and the initialization workflow's existing continuation behavior. It does not introduce retries or require a terminal success frame.

## How was this tested?

Added two regression tests in `tests/cli/test_model_pull.py`, using real HTTPX response objects with mocked HTTP calls and no network access:

- A helper test supplies HTTP 200 with progress followed by an error. It checks that the helper returns `False`, preserves bracket-containing error text, suppresses the success message, and closes the response.
- A CLI test exercises the real helper and checks that the command exits with status 1, displays the error, and does not report success.

Before the fix, both new tests failed for the expected reasons: the helper returned `True` and the CLI exited with status 0. The five existing tests in the file passed.

After the fix:

- `uv run --no-sync pytest tests/cli/test_model_pull.py tests/cli/test_model_cmd.py tests/cli/test_init_guidance.py -q` : **29 passed**.
- `uv run --no-sync pre-commit run --files src/openjarvis/cli/model.py tests/cli/test_model_pull.py` :  **passed** (Ruff and Ruff formatting).
- `uv run --no-sync ruff check src/ tests/` : **passed**.
- `uv run --no-sync ruff format --check src/ tests/` : *passed** (1,383 files already formatted).
- `git diff --check` :  **passed**.

## Checklist

- [ ] Tests pass (`uv run pytest tests/ -v`) - full suite not completed.  29 focused tests passed.
- [ ] Follows registry pattern (if adding new component) - not applicable; no new component.
- [ ] Documentation updated (if applicable) - not applicable


ElliotSlusky (approved):
Reviewed the streamed Ollama error handling and regression coverage. The change correctly detects error frames from HTTP 200 streaming responses, propagates failure to the CLI, preserves literal diagnostics, and closes the response. All required checks pass.

